Issue #13, opened 2015-02-23
PATH pattern does not combine well with comma delimination


alcinnz:
I have a logfile whose lines may include snippets along the line of `a=/some/path, b=some/other/path`. As an ELK user interested in these logs I would like to parse them with LogStash. 

The logstash configuration:

    filter {
      grok {
        match => { "message" => "((a=(?<a>%{PATH})?|b=(?<b>%{PATH})?)(,\s)?)+" }
      }
    }

when given `a=/some/path, b=/some/other/path` LogStash gives the output:
```
{
         "message" => "a=/some/path, b=/some/other/path",
        "@version" => "1",
      "@timestamp" => "2015-02-23T01:57:56.933Z",
            "type" => "stdin",
            "host" => "gallifry",
               "a" => "/some/path,"
}
```
I expect b to bind to `some/other/path`:

```
{
         "message" => "a=/some/path, b=/some/other/path",
        "@version" => "1",
      "@timestamp" => "2015-02-23T01:57:56.933Z",
            "type" => "stdin",
            "host" => "gallifry",
               "a" => "/some/path"
               "b" => "/some/other/path"
}
```

alcinnz:
I find that remove all the `?>` from the UNIX_PATH pattern fixes this issue.

purbon:
Fixed in https://github.com/logstash-plugins/logstash-patterns-core/commit/06de0c1bc566874ff87d9b23765f4fee1b3a573f

PR #14, opened 2015-02-23
Fix: Use greedy matching for UNIXPATH pattern

theflimflam:
If UNIXPATH is set to using the lazy quantifier, then it stops
backtracking correctly when combined with other greedy regexes.

This fixes issue #13


purbon:
This got merged https://github.com/logstash-plugins/logstash-patterns-core/commit/06de0c1bc566874ff87d9b23765f4fee1b3a573f, thanks for your contribution!

PR #15, opened 2015-02-27
HTML5 spec adds some characters to make a valid URI

paulRbr:
for example `<` or `>`

https://html.spec.whatwg.org/.w3c-html-core/urls.html#parsing-urls


purbon:
This got merged https://github.com/logstash-plugins/logstash-patterns-core/commit/ded5d540354478bce90a8263d43a55a034f3faf1

PR #20, opened 2015-03-25
Fix tests, remove coercion on optional fields, update naming to match AWS

DanielRedOak:
Updated s3 and elb field names to match AWS specified names, add matching for TCP ELB logs, remove coercion from optional fields. Test pass with changes

Personally I feel it is best to match the field naming outlined by AWS for the S3 (http://docs.aws.amazon.com/AmazonS3/latest/dev/LogFormat.html) and ELB (http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/access-log-collection.html) formats.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -49,6 +49,7 @@ Contributors:
 * Vincent Bernat (vincentbernat)
 * olagache
 * xiaclo
+* Ryan O'Keeffe (danielredoak)
 
 Note: If you've sent us patches, bug reports, or otherwise contributed to
 Logstash, and you aren't on the list above and want to be, please let us know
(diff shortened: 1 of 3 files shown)


DanielRedOak:
https://github.com/elastic/logstash/issues/2897 seems to indicate that the coercion might be a bug to be fixed

suyograo:
@DanielRedOak can you please sign the CLA? https://github.com/elasticsearch/logstash/blob/master/CONTRIBUTING.md#contribution-steps

DanielRedOak:
Should already be on file actually :)

purbon:
@DanielRedOak sorry I took so long to come back here, I was checking your CLA and is actually signed, but with your personal email, while in the PR you are using another email. To move forward we would need both emails to be the same. Could you make the change somehow? 

Thanks a lot for your understanding,

- purbon

DanielRedOak:
Saw this show up when digging around in my PRs. Is this something that would be valuable still? If so, I can see about cleaning up the CLA business.

kares:
since this is a breaking change and we're moving towards ECS fields, I am closing this one 
(there's similar work for the new set of patterns at https://github.com/logstash-plugins/logstash-patterns-core/pull/287)

PR #22, opened 2015-03-27
Add backslash to URIPARAM

cimnine:
So I had this URL with \u00ab pattern, like "http://f00.ch/?q=Som\u0065+on\u0065"

The corresponding RFC does not explicit mention that backslashes are allowed in the query part, but nor denies it. To me it's like the backslash was forgotten when the spec was written.


purbon:
@cimnine In order to move this change forward you should sign the CLA, more info found in https://www.elastic.co/contributor-agreement ... then we also have a way to add test for patterns, so we can track problems sooner, can you add some regression test for this issue.

cimnine:
@purbon  The CLA has been signed today. I will look into adding tests later.

cimnine:
I'm way to inexperienced with git ^^ let me do this PR right and reopen it once I have tests.

PR #24, opened 2015-04-13
Make pattern specs easier to write

jsvd:
This PR creates a helper that provides the user the following interface:

Putting this in `spec/patterns/core_spec.rb`:

```ruby
grok "IPORHOST", "127.0.0.1"
grok "NAGIOS_TIMEPERIOD_TRANSITION", "TIMEPERIOD TRANSITION: 24x7_sans_holidays;-1;1", {"nagios_type" => "TIMEPERIOD TRANSITION" }
```

and running rspec gives you:

```
% bundle exec rspec --format=documentation
Using Accessor#strict_set for specs
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :broken=>true, :export_cypher=>true, :integration=>true, :windows=>true}

NAGIOS_TIMEPERIOD_TRANSITION
  testing line "TIMEPERIOD TRANSITION: 24x7_sans_holidays;-1;1"
    matches
    has expected fields "{"nagios_type"=>"TIMEPERIOD TRANSITION"}"

IPORHOST
  testing line "127.0.0.1"
    matches

Finished in 0.34 seconds
3 examples, 0 failures
```

I don't consider this a final proposal but an idea to improve grok specs writing.

PR #27, opened 2015-04-18
Provide a simple way to run grok patterns in test

purbon:
This PR provide a set of helpers useful to test grok expressions. It relates to #24, but in this PR the helpers created use the RSpec DSL without interfering with it, so anyone can actually use this helpers in other context if extended.

It also provide an example of simple and basic custom matcher that simplify the process of checking for the outcome of a test example.

### DIFF
--- a/spec/patterns/core_spec.rb
+++ b/spec/patterns/core_spec.rb
@@ -1,6 +1,31 @@
 # encoding: utf-8
-require "logstash/devutils/rspec/spec_helper"
-require 'logstash/patterns/core'
+require "spec_helper"
+require "logstash/patterns/core"
 
-describe LogStash::Patterns::Core do
+describe "SYSLOGLINE" do
+
+  let(:value)   { "Mar 16 00:01:25 evita postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]" }
+  let(:grok)    { grok_match(subject, value) }
+  it "a pattern pass the grok expression" do
+    expect(grok).to pass
+  end
+
+  it "matches a simple message" do
+    expect(subject).to match(value)
+  end
+
+  it "generates the program field" do
+    expect(grok_match(subject, value)).to include("program" => "postfix/smtpd")
+  end
+
+end
+
+describe "COMMONAPACHELOG" do
+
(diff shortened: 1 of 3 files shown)


purbon:
@jsvd could you review and provide feedback to this way of providing helpers? I think this might a more clean way in comparison with the one proposed in #24.

I didn't add here yet the check for fields, but is not something hard to archive in a similar way :-)

jsvd:
+1 to properly extending rspec!

Do you think it's possible to have a more "pattern" oriented spec DSL such as:

```ruby
describe "IPORHOST" do #describe a pattern
  it "matches a simple message" do
    expect(subject).to match("127.0.0.1")
  end
  it "generates field ip" do
    expect(subject.match("127.0.0.1)).to include("ip" => "127.0.0.1")
  end
end
```

purbon:
@jsvd yeah! we can get very close to this. Good idea!

purbon:
@jsvd I added a few of your comments, if you like the way it is now I can add more patterns in here so we can publish this improvement. Do you agree?

spec/patterns/core_spec.rb
```diff
@@ -1,6 +1,40 @@
...
+
+  let(:pattern_name) { "SYSLOGLINE"  }
+  let(:value)        { "Mar 16 00:01:25 evita postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]" }
+  let(:grok)         { build_grok(pattern_name) }
+
+  it "a pattern pass the grok expression" do
+    expect(grok_match(pattern_name, value)).to pass
+  end
+
+  describe "SYSLOGLINE" do
```
jsvd: How about not having a top level describe of Patterns::Core so we can have multiple "describe pattern" blocks that will not need a subject { } ?
like

```ruby
describe "SYSLOGLINE" do

  let(:value) { "Mar 16 00:01:25 evita postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]" }
  it "matches a simple message" do
    expect(subject).to match(value)
  end

  it "generates the program field" do
    expect(grok_match(subject, value)).to include("program" => "postfix/smtpd")
  end
end
```

purbon: sure, the top level thing is completely irrelevant.

jsvd:
LGTM, this is awesome @purbon :dancer:

electrical:
Impressive work @purbon eventually we should have full pattern tests for everything including every field it can create.

purbon:
makes sense @electrical, there are a lot of simple helpers we can add to simplify the way rspec is used ... more in the pipeline.

PR #23, opened 2015-04-13
This pattern does not catch nagios-3.5.1 time transition log

geogdog:
It doesn't include the trailing semi-colon (;)

Examples from my logfile:
```
[1428411344] TIMEPERIOD TRANSITION: 24X7;-1;1
[1428411344] TIMEPERIOD TRANSITION: 24x7;-1;1
[1428411344] TIMEPERIOD TRANSITION: 24x7_sans_holidays;-1;1
[1428411344] TIMEPERIOD TRANSITION: none;-1;0
[1428411344] TIMEPERIOD TRANSITION: us-holidays;-1;0
[1428411344] TIMEPERIOD TRANSITION: workhours;-1;1
[1428411360] TIMEPERIOD TRANSITION: 24X7;-1;1
[1428411360] TIMEPERIOD TRANSITION: 24x7;-1;1
[1428411360] TIMEPERIOD TRANSITION: 24x7_sans_holidays;-1;1
[1428411360] TIMEPERIOD TRANSITION: none;-1;0
[1428411360] TIMEPERIOD TRANSITION: us-holidays;-1;0
[1428411360] TIMEPERIOD TRANSITION: workhours;-1;1
[1428479803] TIMEPERIOD TRANSITION: 24X7;-1;1
[1428479803] TIMEPERIOD TRANSITION: 24x7;-1;1
[1428479803] TIMEPERIOD TRANSITION: 24x7_sans_holidays;-1;1
[1428479803] TIMEPERIOD TRANSITION: none;-1;0
[1428479803] TIMEPERIOD TRANSITION: us-holidays;-1;0
[1428479803] TIMEPERIOD TRANSITION: workhours;-1;1
[1428479808] TIMEPERIOD TRANSITION: 24X7;-1;1
[1428479808] TIMEPERIOD TRANSITION: 24x7;-1;1
[1428479808] TIMEPERIOD TRANSITION: 24x7_sans_holidays;-1;1
[1428479808] TIMEPERIOD TRANSITION: none;-1;0
[1428479808] TIMEPERIOD TRANSITION: us-holidays;-1;0
[1428479808] TIMEPERIOD TRANSITION: workhours;-1;1
[1428482856] TIMEPERIOD TRANSITION: 24X7;-1;1
[1428482856] TIMEPERIOD TRANSITION: 24x7;-1;1
[1428482856] TIMEPERIOD TRANSITION: 24x7_sans_holidays;-1;1
[1428482856] TIMEPERIOD TRANSITION: none;-1;0
[...]

PR #30, opened 2015-04-23
Nagios patterns

geogdog:
Originally logged as #23.  I've added tests and updated the pattern.

### DIFF
--- a/spec/patterns/nagios_spec.rb
+++ b/spec/patterns/nagios_spec.rb
@@ -1,0 +1,198 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/patterns/core"
+
+describe "NAGIOSLOGLINE - CURRENT HOST STATE" do
+
+  let(:value)   { "[1427925600] CURRENT HOST STATE: nagioshost;UP;HARD;1;PING OK - Packet loss = 0%, RTA = 2.24 ms" }
+  let(:grok)    { grok_match(subject, value) }
+
+  it "a pattern pass the grok expression" do
+    expect(grok).to pass
+  end
+
+  it "matches a simple message" do
+    expect(subject).to match(value)
+  end
+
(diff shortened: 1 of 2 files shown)


spec/patterns/nagios_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+  it "a pattern pass the grok expression" do
+    expect(grok).to pass
+  end
+
+  it "matches a simple message" do
+    expect(subject).to match(value)
+  end
+
+  it "generates the nagios_type field" do
+    expect(grok_match(subject, value)).to include("nagios_type" => "SERVICE ALERT")
```
purbon: In here you can use the value created in line 113, no need to call again the pattern matching.

purbon:
Hi, I must admit I'm not an expert on nagios log format. Can you also add test to check if this change still support the old format? this would be very nice.

geogdog:
Hi,

As far I can tell...this format has always been incorrect.  Removing the semi-colon is actually a bug fix.  Maybe I could add a regression test to make sure it doesn't creep back in?

spec/patterns/nagios_spec.rb
geogdog: You mean replace the occurances of:
```
expect(grok_match(subject, value)).to
```
with just:
```
expect(grok).to
```
?

purbon:
Hi,
  I did a search for this kind of logs and I agree this is a bug fix, looks like this has been always there, so now not sure why this got introduced. I think last things to get this merged are:

* Add a regression test.
* Do a minor version bump.
* Rebase the PR.

Looking forward to have this in.

geogdog:
I've rebased against master and added the regression test for the TIMEPERIOD TRANSITION bug.

purbon:
LGTM

PR #36, opened 2015-05-11
Make PROG a bit more flexible

simmel:
This small patch makes logstash correctly interpret e.g. ansible's "ansible-" progname. ATM it failes with a grok parse error.

Generally any characters are allowed except [ and : (colon needed because sometimes a [pid] isn't used).

Moved here from elastic/logstash#1719


jsvd:
can you add a small spec that shows a use case that is improved with this patch? Writing tests for patterns in now much easier. example:

```ruby
describe "SYSLOGLINE" do

  it "matches a simple message" do
    expect(subject).to match("Mar 16 00:01:25 evita postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]")
  end
end
```

simmel:
Like that @jsvd? Maybe we can discuss on IRC how to make the tests better? Also, I *love* that you always give an example when suggesting a change! Make it *much* easier and faster to get into the whole mindset of what you are suggesting. Great artists steal and all that = )

jsvd:
Maybe I'd just add <> the list of acceptable characters. it's odd to reject [ but not ], I think. 
This change also allows spaces to be used in the program and although the [spec actually allows any usascii from 33 to 126](https://tools.ietf.org/html/rfc5424#section-6.2.5) I'm not sure how many popular syslog implementations support that whole character range.

Also, thank you :) testing is still quite difficult on logstash so providing an example is the easiest way to lower that barrier.

simmel:
On Mon, 2015-05-11 at 07:26:49 -0700, João Duarte wrote:
> Maybe I'd just add <> the list of acceptable characters. it's odd to reject [ but not ], I think. 
> This change also allows spaces to be used in the program and although the [spec actually allows any usascii from 33 to 126](https://tools.ietf.org/html/rfc5424#section-6.2.5) I'm not sure how many popular syslog implementations support that whole character range.

Let's just go all in then? = ) New code supports RFC5424's APP-NAME
fully (except [ and ] of course since we need to match the eventual pid
with that). I also removed the non-capturing grouping since I didn't see
why it was in use anyway.

> Also, thank you :) testing is still quite difficult on logstash so providing an example is the easiest way to lower that barrier.

Well, you all have done a tremendous jobs improving that situation!

jsvd:
LGTM

jsvd:
@simmel could you do step 2 of the [contributors guide](https://github.com/elastic/logstash/blob/master/CONTRIBUTING.md#contribution-steps)?

simmel:
My boss (Mattias Nordström) has already signed the CLA for us (Stockholm
University) on 2015-02-17 and I was added to the "Schedule A" appendix.

Do you need our transaction ID?

jsvd:
@simmel sorry about the bureaucracy but is it possible for you to resubmit the patch with the email under which you were attached with to the cla?

simmel:
Ah, of course! That was my bad, sorry! It should be fixed now. My work should be credited for having me improve logstash during working hours = )

PR #44, opened 2015-05-21
Added grok parser for bro http.log

fedelemantuano:
Bro (https://www.bro.org/index.html) parser for http.log.


purbon:
@fedelemantuano we've in place a way to test the patterns, can you add test for this change? You can take inspiration from https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/nagios_spec.rb for example. Let me know if you need more guidance.

fedelemantuano:
@purbon I added bro http.log tester. I took inspiration from nagios_spec.rb.

patterns/bro
```diff
@@ -0,0 +1,2 @@
+# http.log (https://www.bro.org/sphinx/scripts/base/protocols/http/main.bro.html#type-HTTP::Info)
+HTTP %{NUMBER:ts}\t%{NOTSPACE:uid}\t%{IP:orig_h}\t%{INT:orig_p}\t%{IP:resp_h}\t%{INT:resp_p}\t%{INT:trans_depth}\t%{GREEDYDATA:method}\t%{GREEDYDATA:domain}\t%{GREEDYDATA:uri}\t%{GREEDYDATA:referrer}\t%{GREEDYDATA:user_agent}\t%{NUMBER:request_body_len}\t%{NUMBER:response_body_len}\t%{GREEDYDATA:status_code}\t%{GREEDYDATA:status_msg}\t%{GREEDYDATA:info_code}\t%{GREEDYDATA:info_msg}\t%{GREEDYDATA:filename}\t%{GREEDYDATA:bro_tags}\t%{GREEDYDATA:username}\t%{GREEDYDATA:password}\t%{GREEDYDATA:proxied}\t%{GREEDYDATA:orig_fuids}\t%{GREEDYDATA:orig_mime_types}\t%{GREEDYDATA:resp_fuids}\t%{GREEDYDATA:resp_mime_types}
```
purbon: I have a couple of things here:

* I think naming this pattern ```BRO_HTTP_INFO``` makes more sense to me than just HTTP, then people can namespace this pattern better, makes sense to you?
* (I don't know this kind of logs, so please educate me), is there always only one tab between fields?, is this separator always a tab?

purbon:
I just left one small comment, for the rest it looks good to me for now. 

BTW @fedelemantuano, do you have more examples of this logs? are all the fields always there?

Thanks for your contribution,

patterns/bro
fedelemantuano: BRO_HTTP or BRO_HTTP_INFO it's better than only HTTP, so for me it's ok.
There is always tab between fields.
I can generate some log from my navigation and share them with you.

fedelemantuano:
Hi @purbon I sent you some examples of bro logs in my shared gdrive folder.

patterns/bro
jordansissel: This tab-separated format is probably best handled by the CSV filter, not
grok. Thoughts?

fedelemantuano: Yes the csv filter can be a solution, every bro log must have a own custom "columns" like this:

 __columns => ["ts","uid","id.orig_h","id.orig_p","id.resp_h","id.resp_p",...__

custom separator (\t) and void quote_char. I tested with __quote_char => ""__ and  __quote_char => " "__, but this solution doesn't work.

So I thought to do a custom parser with correct name for every bro logs.

PR #45, opened 2015-05-22
Apache error log format

chernjie:
* Add apache error log formats

### DIFF
--- a/patterns/grok-patterns
+++ b/patterns/grok-patterns
@@ -89,6 +89,8 @@ QS %{QUOTEDSTRING}
 SYSLOGBASE %{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{SYSLOGPROG}:
 COMMONAPACHELOG %{IPORHOST:clientip} %{USER:ident} %{USER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
 COMBINEDAPACHELOG %{COMMONAPACHELOG} %{QS:referrer} %{QS:agent}
+APACHEERRORTIME %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
+APACHEERRORLOG \[%{APACHEERRORTIME:timestamp}\] \[%{LOGLEVEL:loglevel}\] (?:\[%{WORD:originator} %{IPORHOST:clientip}\] ){0,1}%{GREEDYDATA:errormsg}
 
 # Log Levels
 LOGLEVEL ([Aa]lert|ALERT|[Tt]race|TRACE|[Dd]ebug|DEBUG|[Nn]otice|NOTICE|[Ii]nfo|INFO|[Ww]arn?(?:ing)?|WARN?(?:ING)?|[Ee]rr?(?:or)?|ERR?(?:OR)?|[Cc]rit?(?:ical)?|CRIT?(?:ICAL)?|[Ff]atal|FATAL|[Ss]evere|SEVERE|EMERG(?:ENCY)?|[Ee]merg(?:ency)?)
--- a/spec/patterns/core_spec.rb
+++ b/spec/patterns/core_spec.rb
@@ -32,7 +32,7 @@ end
 
 describe "APACHEERRORLOG" do
 
-  let(:value) { "[Thu May 01 02:23:39 2014] [error] [client 127.0.0.1] PHP Warning:  phpinfo(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in /var/www/html/index.php on line 1"}
+  let(:value) { "[Thu May 01 02:23:39 2014] [error] [client 127.0.0.1] File does not exist: /var/www/html/404.html" }
 
   it "generates the clientip field" do
     expect(grok_match(subject, value)).to include("clientip" => "127.0.0.1", "timestamp" => "Thu May 01 02:23:39 2014", "loglevel" => "error")
--- a/spec/patterns/core_spec.rb
+++ b/spec/patterns/core_spec.rb
@@ -29,3 +29,13 @@ describe "COMMONAPACHELOG" do
   end
 
 end
+
+describe "APACHEERRORLOG" do
+
+  let(:value) { "[Thu May 01 02:23:39 2014] [error] [client 127.0.0.1] PHP Warning:  phpinfo(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in /var/www/html/index.php on line 1"}
+
+  it "generates the clientip field" do
+    expect(grok_match(subject, value)).to include("clientip" => "127.0.0.1", "timestamp" => "Thu May 01 02:23:39 2014", "loglevel" => "error")
+  end
+
+end


patterns/grok-patterns
```diff
@@ -88,7 +88,9 @@ QS %{QUOTEDSTRING}
 # Log formats
 SYSLOGBASE %{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{SYSLOGPROG}:
 COMMONAPACHELOG %{IPORHOST:clientip} %{USER:ident} %{USER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
-COMBINEDAPACHELOG %{COMMONAPACHELOG} %{QS:referrer} %{QS:agent}
+COMBINEDAPACHELOG %{COMMONAPACHELOG} "%{QS:referrer}" "%{QS:agent}"
```
untergeek: You don't need the quotes around this.  QS refers to "Quoted String," and it is meant to keep the quotes around it.

Changing this would break configurations for a great many people.  The correct way to strip quotes from around a quoted string would be to use mutate gsub.

chernjie: Thanks @untergeek

untergeek:
Sorry to keep you waiting.  Is there a chance you could add a test to https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/core_spec.rb for this?

Mister-X-:
I got different 3 different formats in that file (I'm using Phusion Passenger for Snorby on Ubuntu 14.04) and you can see 2 of these formats here (third one is just messages with no time/date):

```
[ 2015-05-25 06:48:46.2427 43183/7fe55537f7c0 age/Wat/Main.cpp:1264 ]: Starting PassengerAgent watchdog...
[ 2015-05-25 06:48:46.2473 43186/7f0fde2de7c0 age/Hel/Main.cpp:944 ]: Starting PassengerAgent server...
[ 2015-05-25 06:48:46.2598 43186/7f0fde2de7c0 age/Hel/Main.cpp:237 ]: PassengerAgent server running in multi-application mode.
[ 2015-05-25 06:48:46.2613 43186/7f0fde2de7c0 age/Hel/Main.cpp:698 ]: PassengerAgent server online, PID 43186
[ 2015-05-25 06:48:46.2667 43195/7fb47b2f17c0 age/Log/Main.cpp:442 ]: Starting PassengerAgent logger...
[ 2015-05-25 06:48:46.2672 43195/7fb47b2f17c0 age/Log/Main.cpp:323 ]: PassengerAgent logger online, PID 43195
[Mon May 25 06:48:46.268033 2015] [ssl:error] [pid 2023:tid 140163329025920] AH02217: ssl_stapling_init_cert: can't retrieve issuer certificate! [subject: emailAddress=support@company.com,CN=CommonNameSomething,O=Something LLC,L=Bleh,ST=California,C=US / issuer: emailAddress=support@company.com,CN=CommonNameSomething,O=Something LLC,L=Bleh,ST=California,C=US / serial: AACBB16685C6710A / notbefore: Apr  6 17:01:25 2015 GMT / notafter: Apr  6 17:01:25 2018 GMT]
[Mon May 25 06:48:46.268046 2015] [ssl:error] [pid 2023:tid 140163329025920] AH02235: Unable to configure server certificate for stapling
[Mon May 25 06:48:46.268053 2015] [ssl:warn] [pid 2023:tid 140163329025920] AH01909: RSA certificate configured for Blah:443 does NOT include an ID which matches the server name
[Mon May 25 06:48:46.268194 2015] [mpm_event:notice] [pid 2023:tid 140163329025920] AH00489: Apache/2.4.7 (Ubuntu) Phusion_Passenger/5.0.7 OpenSSL/1.0.1f configured -- resuming normal operations
[Mon May 25 06:48:46.268203 2015] [core:notice] [pid 2023:tid 140163329025920] AH00094: Command line: '/usr/sbin/apache2'
```

purbon:
@Mister-X- are this logs being generated using Apache with Passenger?, can you elaborate more the differences you see with the proposed pattern?

Mister-X-:
Yes, it was generated using Apache with Passenger on Ubuntu 14.04.2 (in this case, using Snorby).

The time format is different. One has brackets before and after the time and the other has more information between the brackets. 

I guess the meaning of the first item is the PID of passenger (you can see it on the fourth line) and I would say the item after the slash is the memory location where that thing is logged. Then, file name and line number.

For the second part, you can get the 'section' and loglevel in the second item in brackets, the pid and tid in the third, then the apache error message.

Is it the answer you're looking for?

purbon:
So as I understand this are the logs generated by the ```passenger_module```, I kinda see your pattern more as a new pattern for passenger, don't you think?

chernjie:
According to [apache man page](http://httpd.apache.org/docs/2.2/logs.html#errorlog), the default apache `ErrorLog` format is 
```
[Wed Oct 11 14:32:52 2000] [error] [client 127.0.0.1] client denied by server configuration: /export/home/live/ap/htdocs/test
```

Would it help if I update the Grok-pattern key with a `COMMON`, e.g. `COMMONAPACHEERRORLOG` instead of `APACHEERRORLOG`

purbon:
makes sense to me @chernjie, then also @Mister-X- can add the ```PASSENGER_APACHE_LOG```.

Mister-X-:
So, you want me to build the pattern for passenger logs? If yes, how should the hex field after the pid called?

purbon:
@Mister-X- I don't know, do you know what It means?

Mister-X-:
I will see if I can create a PR and let you know. If not, I can definitely give you the grok parser for that.

I have no idea what is the meaning of the thing after the pid (the hex value). I'm asking what it is and I will hopefully get an answer soon.

Mister-X-:
It's the thread ID.

purbon:
Hi @chernjie  , thanks a lot for your contribution. In order to move forward with your contribution you will need to sign the CLA, more info at -> https://www.elastic.co/contributor-agreement

purbon:
@Mister-X- where you able to create a new PR? that would be super nice.

purbon:
Closing this for now as the CLA request got stale without an answer for long time, if you change your mind and sign the CLA we would be more than welcome to get this in.

Thanks a lot for your contribution.

purbon:
@chernjie Thanks for signing the CLA, much appreciate :+1:

patterns/grok-patterns
```diff
@@ -89,6 +89,8 @@ QS %{QUOTEDSTRING}
 SYSLOGBASE %{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{SYSLOGPROG}:
 COMMONAPACHELOG %{IPORHOST:clientip} %{USER:ident} %{USER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
 COMBINEDAPACHELOG %{COMMONAPACHELOG} %{QS:referrer} %{QS:agent}
+APACHEERRORTIME %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
```
purbon: COMMONAPACHELOG uses HTTPDATE for their timestamp, this pattern is very similar to your APACHEERRORTIME, why not reusing it? probably making the ```/``` optional? what do you think?

purbon:
@chernjie I think renaming to COMMONAPACHEERRORLOG would be nice.

After that two very small changes I think this is perfect to merge. Do you mind something this very small changes?

/cheers

chernjie:
Apparently someone added 585dae7

purbon:
@chernjie yes, we should normalize this.

chernjie:
Verified that this has been reimplemented in 585dae7 by @igalic two weeks ago, closing this pull request.

patterns/grok-patterns
chernjie: `HTTPDATE` does not satisfy error log timestamp, interestingly 585dae7 has a new pattern called `HTTPERRORDATE` which is exactly the same as `APACHEERRORTIME` :)

chernjie:
@purbon `HTTPD_ERRORLOG` has been released in v0.4.0, if you want to rename it to `COMMONAPACHEERRORLOG` it would break v0.4.0 :)

Otherwise, the new patterns from 585dae7 matches more HTTPD variations than the pattern I have above. With the exception of "client" which is hardcoded in 585dae7.

If you are fine with the above two, I think there is nothing else to normalize. :)

purbon:
@chernjie I think we should rename to a ```COMMONAPACHEERRORLOG``` sooner or later, when I released ```v0.4.0``` I had this issue not in my mind. Then we can release ```v1.0.0```, what do you think?

chernjie:
@purbon I think `HTTPD_ERRORLOG` is better, and would suggest changing the following instead:
- `COMMONAPACHELOG` to `HTTPD_COMMONLOG`
- `COMBINEDAPACHELOG` to `HTTPD_COMBINEDLOG`

One more suggestion is to move all `APACHE|HTTPD*` into it's own file `patterns/httpd`.

Of course, if we need to we can consider aliasing with `s/HTTPD/APACHE/g`

Perhaps we should also get feedback from @igalic?

PR #59, opened 2015-06-10
fix for ISO8601 parsing

bastelfreak:
ISO8601 defines the format without a whitespace. For more information,
take a look at: http://en.wikipedia.org/wiki/ISO_8601


purbon:
Good morning @bastelfreak , thanks a lot for your contribution. In order to move forward with your contribution you will need to sign the CLA, more info at -> https://www.elastic.co/contributor-agreement

On the other side, we just added the option to test the grok patterns we've in here, adding test to validate your changes it will be cool to merge this in too. You can take inspiration to write the test from (https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/syslog_spec.rb), don't heasitate to ask any question you might have, we'll be happy to provide help.

bastelfreak:
signed the CLA, I will take a look at the tests later and let you know if I have any questions.

wiibaa:
Considering the 2 facts that 
* some major few logging library (log4J, logback...) use a space in their "ISO8601"
* this is a strongly confusing topic in logstash community as discussed in https://github.com/elastic/logstash/pull/769

I'm not in favor of this PR as is, but maybe it is time to introduce **ISO8601_STRICT** pattern as proposed in https://github.com/elastic/logstash/pull/769#issuecomment-50217409

purbon:
@wiibaa thank you for your insides, although I'm not sure I understand what you mean in your first point.  You say Log4j, and others, use an space as proposed in here? I did a quick google and it return this page https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/helpers/ISO8601DateFormat.html where the format even have no ```T````. Can you educate me/us a bit more on the topic?

After checking the standard  (http://metric1.org/8601.pdf), there is a note that the ```T``` can be omitted if boths parts agreed, so there is no such a strict version vs relaxed one, boths are perfectly valid. Not sure what to think.

I guess my idea would be to make the ```T``` optional, so it can be a ```T``` or and space, that do you think?

patterns/grok-patterns
```diff
@@ -65,7 +65,7 @@ DATE_US %{MONTHNUM}[/-]%{MONTHDAY}[/-]%{YEAR}
 DATE_EU %{MONTHDAY}[./-]%{MONTHNUM}[./-]%{YEAR}
 ISO8601_TIMEZONE (?:Z|[+-]%{HOUR}(?::?%{MINUTE}))
 ISO8601_SECOND (?:%{SECOND}|60)
-TIMESTAMP_ISO8601 %{YEAR}-%{MONTHNUM}-%{MONTHDAY}[T ]%{HOUR}:?%{MINUTE}(?::?%{SECOND})?%{ISO8601_TIMEZONE}?
+TIMESTAMP_ISO8601 %{YEAR}-%{MONTHNUM}-%{MONTHDAY}[T]%{HOUR}:?%{MINUTE}(?::?%{SECOND})?%{ISO8601_TIMEZONE}?
```
purbon: As checked in the standard (http://metric1.org/8601.pdf), boths separators are perfectly valid. I think the right solution here is to have either a T or an space, but only one of them. What do you think?

purbon:
@bastelfreak I was checking your CLA, and notice you might have used a different email than the one in your commits? is this true? can you verify that? we require the email used to sign it to be the same as the one used in the signature.

wiibaa:
@purbon a good example in the java world is in joda time where the **T** is mandatory in ISO8601 dateTime parser (see http://joda-time.sourceforge.net/apidocs/org/joda/time/format/ISODateTimeFormat.html#dateTimeParser())
vs the **no T** version as you can find it into log4j, logback as expressed in http://jira.qos.ch/browse/LOGBACK-262

I suspect the root cause to be more people (**I included**) using wikipedia as a source of thruth instead of the ISO document. thanks for pointing me to it

purbon:
@wiibaa wikipedia also say T can be avoid (by agreement), but it say it so small :smile_cat: jejjeje, so do you agree on having T or space ? as the standard say?

wiibaa:
@purbon I'm lax on datetime format since a long-time :D
The question must go to this PR author => @bastelfreak thoughts ?

purbon:
@bastelfreak thoughts? 
@wiibaa I can also ask you jejejejej! your opinions as very welcome.

wiibaa:
@purbon my opinion is the same as from https://github.com/elastic/logstash/pull/769 so I would leave this pattern as it is and close this PR. We already have a working solution for all use case even the corner-case of implementation details vs standards specs so :see_no_evil:

purbon:
Closing this PR as of already having a working solution, we can reopen if we change our mind on how strict do we want to be with the ISO and to include it here if necessary.

On the other side, CLA request is still stale since some time.

If this change we can revisit for sure.

/cheers

PR #61, opened 2015-06-24
make field progname optional as it is not always supplied

csamsel:
I am receiving messages like the following from a synology nas device: 

<14>Jun 24 10:32:02 hostname WinFileService Event: read, Path: /.DS_Store, File/Folder: File, Size: 6.00 KB, User: user@host, IP: 123.123.123.123

as you can see they do not contain the program part of the syslog message, so i made it optional in the parser. I dont see any disadvantages but i am not sure if the syntax is optimal.

### DIFF
--- a/patterns/linux-syslog
+++ b/patterns/linux-syslog
@@ -1,6 +1,6 @@
 SYSLOG5424PRINTASCII [!-~]+
 
-SYSLOGBASE2 (?:%{SYSLOGTIMESTAMP:timestamp}|%{TIMESTAMP_ISO8601:timestamp8601}) (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{SYSLOGPROG}:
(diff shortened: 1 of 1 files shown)


purbon:
@csamsel Thanks a lot for your contribution.

To move forward, it would be nice if you can add some test to verify your changes are working as expected. To write the test you can fetch inspiration from (https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/syslog_spec.rb), don't heasitate to ask any question you might have, we'll be more than welcome to help.

purbon:
This got merge in https://github.com/logstash-plugins/logstash-patterns-core/commit/b58ab94ffe88b8c32d67eedb3a692a722486f008 with test added by myself in https://github.com/logstash-plugins/logstash-patterns-core/commit/969cc77d10a71ddf911d87fc1294e571100cad65

Thanks for your contribution

PR #63, opened 2015-06-26
Changed IPV4 address regex for possible improvement

akshat1991:
This is my first patch to any open source project, so if there is some blunder I made, please guide me.
I found that the current IPV4 address pattern checks for 250-255 first, then 200-249 and finally 1-199. Statistically, most octets would fall in the last range, due to more numbers and private IP patterns(10.* and 192.*) so it makes sense to check in the reverse order. 
In regex engine, for some text parsing jobs, I found about 10% improvement in performance.
I know this is a micro-improvement(which was important as it was a bottleneck in my project), but since it is not hurting readability, I thought I'll just try to contribute.


akshat1991:
I have signed the CLA for elastic search. Also the link for details points to https://github.com/logstash-plugins/logstash-patterns-core/blob/master/CONTRIBUTING.md#contribution-steps. The file is not present.

purbon:
@akshat1991 you are right, this file in this missing in this repo. As it's in all Logstash plugins, every pointer to a file like this should be getting you to the file with the same name at the main logstash core repo (https://github.com/elastic/logstash/blob/master/CONTRIBUTING.md). Can you educate me where you found this link?

purbon:
Speaking about the change, it might makes sense (still need to run some microbench to see the speedup). However to move forward it would be nice if you can add some test to verify your changes are working as expected. To write the test you can fetch inspiration from (https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/syslog_spec.rb), don't heasitate to ask any question you might have, we'll be more than welcome to help.

purbon:
This got merged https://github.com/logstash-plugins/logstash-patterns-core/commit/9381b0beb4f0a00bec35a89c09254bd799d093c1 with some basic test, but still cool to see the benchmarks @akshat1991 .

Thanks a lot for your contribution

PR #69, opened 2015-07-17
factor out haproxy groking to be used outside of syslog parsing

human39:
If the event was previously parsed by the syslog input the syslog fields are no longer present.   In this case, we need a pattern that doesn't expect the message to be a full  syslog message.


patterns/haproxy
```diff
@@ -31,7 +31,9 @@ HAPROXYCAPTUREDRESPONSEHEADERS %{DATA:captured_response_headers}
...
+HAPROXYHTTPBASE %{IP:client_ip}:%{INT:client_port} \[%{HAPROXYDATE:accept_date}\] %{NOTSPACE:frontend_name} %{NOTSPACE:backend_name}/%{NOTSPACE:server_name} %{INT:time_request}/%{INT:time_queue}/%{INT:time_backend_connect}/%{INT:time_backend_response}/%{NOTSPACE:time_duration} %{INT:http_status_code} %{NOTSPACE:bytes_read} %{DATA:captured_request_cookie} %{DATA:captured_response_cookie} %{NOTSPACE:termination_state} %{INT:actconn}/%{INT:feconn}/%{INT:beconn}/%{INT:srvconn}/%{NOTSPACE:retries} %{INT:srv_queue}/%{INT:backend_queue} (\{%{HAPROXYCAPTUREDREQUESTHEADERS}\})?( )?(\{%{HAPROXYCAPTUREDRESPONSEHEADERS}\})?( )?"(<BADREQ>|(%{WORD:http_verb} (%{URIPROTO:http_proto}://)?(?:%{USER:http_user}(?::[^@]*)?@)?(?:%{URIHOST:http_host})?(?:%{URIPATHPARAM:http_request})?( HTTP/%{NUMBER:http_version})?))?"
```
purbon: I wonder if BASE is a good name for this pattern? what do you think?

purbon:
Thanks for your contribution @human39, nowadays we added test to this set of patterns. In order to get this in, it would be great if you could add a few test to validate this pattern works as expected. To write the test you can take inspiration from https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/syslog_spec.rb (for example), please don't heasitate to ask any questions  you might have, we'll be happy to provide help.

patterns/haproxy
human39: I didn't want to change the 'HAPROXYHTTP' value here because of the possibility of breaking currently deployed filter rules.  I didn't put a lot of thought into BASE and I'm open to suggestions.  I couldn't find a standard naming convention to mimic.

human39:
@purbon I committed tests as requested.  As stated previously, I'm not married to the variable names including 'BASE'.  We can discuss what you would want instead.

Thanks!

patterns/haproxy
purbon: Makes sense, I did a quick check and was not really able to figure out a "cool" name for this. I'm also terrible with naming. We can leave it like this for now and later on if desired we can change it.

purbon:
Hi, code LGTM,  I don't know why, but yesterday I forgot to mention that before we can move forward it's necessary to sign the CLA https://www.elastic.co/contributor-agreement. I apologies for the inconvenience.

Looking forward to get this in.

- purbon

human39:
@purbon - CLA signed.  Please let me know if there are any issues.

PR #72, opened 2015-08-04
add a pattern for rails 3

igalic:
This is the rails pattern we use in our configuration.
It's based off @mudge's [gist](https://gist.github.com/mudge/5047696) and @ese's improvements.

### DIFF
--- a/patterns/rails
+++ b/patterns/rails
@@ -1,0 +1,13 @@
+RUUID \h{32}
+# rails controller with action
+RCONTROLLER (?<controller>[^#]+)#(?<action>\w+)
+
+# this will often be the only line:
+RAILS3HEAD (?m)Started %{WORD:verb} "%{URIPATHPARAM:request}" for %{IPORHOST:clientip} at (?<timestamp>%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND} %{ISO8601_TIMEZONE})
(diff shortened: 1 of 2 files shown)


igalic:
re tests (since most of the other pull requests here are labeled with missing them)

```
igalic@levix ~/src/elastic/logstash-patterns-core (git)-[patch-1] % bundle
Your Gemfile lists the gem logstash-filter-grok (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
Fetching gem metadata from https://rubygems.org/..........
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Could not find gem 'logstash-devutils (>= 0) ruby' in any of the gem sources listed in your Gemfile or available on this machine.
```

ph:
@igalic You have to run JRuby to be able to run the tests, if you are using rbenv or rvm it is a simple install.

igalic:
@ph thanks for the pointer.
did that just now, and can run tests just fine!
yay! partial success. now i have to figure out how to spec test multi-line patterns.
(as soon as i do, i should provide a patch for the java patterns as well ;)

*****

@ese please note that we have largely no influence on these applications (gitlab, redmine)

igalic:
@ph added two tests

purbon:
LGTM

PR #81, opened 2015-08-27
accomodate '=' sign in file names.

mnikhil-git:


talevy:
thanks for the addition!

would you mind signing our Contributor License Agreement by following step 2 found here: https://github.com/elastic/logstash/blob/master/CONTRIBUTING.md#contribution-steps.

Also, it would be great to have tests with this change. You may want to look at some of the existing specs to see how that can be achieved.

thanks!

purbon:
Hi, in order to check for this PR I added a simple test https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/core_spec.rb#L97 for the path. If you rebase this PR with master, you will see how this does not pass.

Thanks a lot for your contribution,

- purbon

patterns/grok-patterns
```diff
@@ -32,6 +32,7 @@ HOSTPORT %{IPORHOST}:%{POSINT}
 # paths
 PATH (?:%{UNIXPATH}|%{WINPATH})
 UNIXPATH (?>/(?>[\w_%!$@:.,~-]+|\\.)*)+
+UNIXPATH (?>/(?>[\w_%!$@:.,~-=]+|\\.)*)+
```
purbon: why don't you simply add the new character in the old path? this should be there. Duplicated paths would cause problems here.

jordansissel: Agreed, looks like the previous definition of UNIXPATH was left accidentally.

talevy:
Closing due to needed followup that has potentially gone stale. feel free to re-open!

PR #93, opened 2015-09-08
httpd: sync names between httpd20 and httpd24

igalic:
make sure names between httpd 2.0 and httpd 2.4 are more or less close to each other


purbon:
@igalic thanks a lot for your contribution, would you be nice to update the test accordingly ? thanks a lot. Tests are found https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/httpd_spec.rb

igalic:
@purbon that spec only tests for a match, not what, exactly, that match contains, but you're right, maybe i should've written those tests betterer to begin with :_)

purbon:
@igalic take the chance to improve the test with the new names, this would be super awesome!

Issue #83, opened 2015-09-01
Make pattern elements matching integers to emit integer fields


magnusbaeck:
There are a number of patterns that match integers (or floats) via e.g. `%{INT:foo}` that are emitting string values for values that cannot be anything but numeric. This is an annoyance since it forces users to define their own Elasticsearch index templates with explicit mappings to get the fields correctly mapped in Elasticsearch. Users shouldn't have to do that if all they want to do is parse and visualize an Apache log; index templates should be for experienced users.

Example with problematic tokens highlighted:

HAPROXYHTTP %{SYSLOGTIMESTAMP:syslog_timestamp} %{IPORHOST:syslog_server} %{SYSLOGPROG}: %{IP:client_ip}:**%{INT:client_port}** \[%{HAPROXYDATE:accept_date}\] %{NOTSPACE:frontend_name} %{NOTSPACE:backend_name}/%{NOTSPACE:server_name} **%{INT:time_request}**/**%{INT:time_queue}**/**%{INT:time_backend_connect}**/**%{INT:time_backend_response}**/%{NOTSPACE:time_duration} %{INT:http_status_code} %{NOTSPACE:bytes_read} %{DATA:captured_request_cookie} %{DATA:captured_response_cookie} %{NOTSPACE:termination_state} **%{INT:actconn}**/**%{INT:feconn}**/**%{INT:beconn}**/**%{INT:srvconn}**/%{NOTSPACE:retries} **%{INT:srv_queue}**/**%{INT:backend_queue}** (\{%{HAPROXYCAPTUREDREQUESTHEADERS}\})?( )?(\{%{HAPROXYCAPTUREDRESPONSEHEADERS}\})?( )?"(<BADREQ>|(%{WORD:http_verb} (%{URIPROTO:http_proto}://)?(?:%{USER:http_user}(?::[^@]*)?@)?(?:%{URIHOST:http_host})?(?:%{URIPATHPARAM:http_request})?( HTTP/%{NUMBER:http_version})?))?"

jordansissel:
Agreed we should have better defaults on built-in patterns that number/int/etc are represented by the appropriate types internally. However, changing this will break backwards compatibility, so we'll have to either be very careful with announcements _or_ make new patterns (HAPROXYHTTP2)

magnusbaeck:
Yep. While I don't like the idea of increasing the number of patterns it _would_ be a backwards compatibility break. I doubt that many are relying on the fields to be strings, though.

[...]

PR #102, opened 2015-09-14
Normalize HTTPD and APACHE

chernjie:


chernjie:
@purbon it should not break compatibility now

patterns/httpd
```diff
@@ -0,0 +1,15 @@
+HTTPDUSER %{EMAILADDRESS}|%{USER}
+HTTPDERROR_DATE %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
+
+# Log formats
+HTTPD_COMMONLOG %{IPORHOST:clientip} %{HTTPDUSER:ident} %{USER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
```
jordansissel: Oh lovely!

While we're here, any thoughts on making `bytes` and `httpversion` both numbers? (int and float, in grok, respectively)

spec/patterns/httpd_spec.rb
```diff
@@ -2,23 +2,112 @@
 require "spec_helper"
 require "logstash/patterns/core"
 
+describe "HTTPD_COMBINEDLOG" do
+
+  let(:value) { '83.149.9.216 - - [24/Feb/2015:23:13:42 +0000] "GET /presentations/logstash-monitorama-2013/images/kibana-search.png HTTP/1.1" 200 203023 "http://semicomplete.com/presentations/logstash-monitorama-2013/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/32.0.1700.77 Safari/537.36"'}
+
+  it "generates the clientip field" do
+    expect(grok_match(subject, value)).to include(
+      'clientip' => '83.149.9.216',
```
jordansissel: Whoa, this works? Amazing! I love rspec so much. Thanks for teaching me this :)

patterns/httpd
jordansissel: If we do change, for example, to use `%{NUMBER:bytes:int}` we'll want the specs to expect numbers, not strings.

For background, I've wanted bytes (especially) and httpversion stored as numbers in Elasticsearch for years, but I didn't want to break compat. This PR is a perfect opportunity to do that.

Thoughts?

chernjie: it makes sense to make bytes `int`, then you can plot bandwidth in Kibana.
As for `httpversion`, care to enlighten me what are the benefits?

purbon: @jordansissel I think we should convert more of this strings to integers,  see #83, I think it makes sense to to do it everywhere that makes sense in the patterns. What do you think?

spec/patterns/httpd_spec.rb
purbon: ++ This works and is super nice! More intel: https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/include-matcher

chernjie:
hi @jordansissel  and @purbon , any updates for this PR?

patterns/httpd
```diff
@@ -0,0 +1,15 @@
+HTTPDUSER %{EMAILADDRESS}|%{USER}
+HTTPDERROR_DATE %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
+
+# Log formats
+HTTPD_COMMONLOG %{IPORHOST:clientip} %{HTTPDUSER:ident} %{USER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes:int}|-)
```
jordansissel: Should httpversion be `:float` also? Or no?

chernjie: well, no harm future-proofing. As of today, I only know of `1.0` and `1.1` in the wild. `0.9` is merged into `1.0` and `2.0` is not commonly used yet.

chernjie: Do you want `response` to be `%{NUMBER:response:int}`

chernjie: @jordansissel any updates on this?

jsvd:
jenkins, test it

joshuaspence:
+1 on this change.

jsvd:
@chernjie could you rebase this against master so I can merge it? thanks!

chernjie:
@jsvd rebased. Should I separate the "breaking-compatibility" commits into a separate pull request?

jsvd:
Does it still break compatibility? seems to maintain all previous patterns

chernjie:
the last 3 commits break compatibility, changing:
- bytes to `int`
- httpversion to `float`
- response to `int`

chernjie:
I think it would be better if the last 3 changes are moved into a separate PR

jsvd:
+1

chernjie:
@jsvd last three commits have been moved to #174

jsvd:
LGTM

jsvd:
great work @chernjie! thanks for the patience.

Issue #73, opened 2015-08-13
AWS ELB log regex doesn't support SSL cipher/protocol


aodj:
The [AWS ELB format spec](https://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/access-log-collection.html) includes SSL cipher and SSL protocol fields that aren't parsed by the currently available `ELB_ACCESS_LOG` regex.

purbon:
Thanks for your feedback @aodj, are you willing to contribute this patterns back? Do you know/have example logs to write the test for this patterns? I guess we might also use the ones in the spec.

 /cheers

salexpdx:
I came across the same issue this evening and am working on a fix for it tonight

salexpdx:
Hi @purbon, I filled out the CLA a few days ago but haven't seen this get merged so wanted to post a comment here letting you know it was done.

kares:
this should be addressed with https://github.com/logstash-plugins/logstash-patterns-core/pull/287, if we missed something please open a new issue with the specifics

PR #104, opened 2015-09-15
Issue73 awselb fix

salexpdx:
Issue #73 is about the three missing fields in the aws ELB logs.  This grabs the fields and brings them in as agent, cipher, and protocol.  This is slightly different than what I have in my LS installation so please review carefully.

### DIFF
--- a/patterns/aws
+++ b/patterns/aws
@@ -8,4 +8,4 @@ ELB_URI %{URIPROTO:proto}://(?:%{USER}(?::[^@]*)?@)?(?:%{URIHOST:urihost})?(?:%{
 
 ELB_REQUEST_LINE (?:%{WORD:verb} %{ELB_URI:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})
 
-ELB_ACCESS_LOG %{TIMESTAMP_ISO8601:timestamp} %{NOTSPACE:elb} %{IP:clientip}:%{INT:clientport:int} (?:(%{IP:backendip}:?:%{INT:backendport:int})|-) %{NUMBER:request_processing_time:float} %{NUMBER:backend_processing_time:float} %{NUMBER:response_processing_time:float} %{INT:response:int} %{INT:backend_response:int} %{INT:received_bytes:int} %{INT:bytes:int} "%{ELB_REQUEST_LINE}" %{QS:agent} %{NOTSPACE:cipher} %{NOTSPACE:protocol} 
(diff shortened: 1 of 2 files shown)


purbon:
Thanks a lot for your contribution @salexpdx !! before we can move forward it would be required to sign the CLA https://www.elastic.co/contributor-agreement. For this PR you used this commit details ```Scott Alexander <github@tagless.com>``` when signing the CLA it would be necessary to use this same contact details.

Looking forward to it,

patterns/aws
```diff
@@ -8,4 +8,4 @@ ELB_URI %{URIPROTO:proto}://(?:%{USER}(?::[^@]*)?@)?(?:%{URIHOST:urihost})?(?:%{
 
 ELB_REQUEST_LINE (?:%{WORD:verb} %{ELB_URI:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})
 
-ELB_ACCESS_LOG %{TIMESTAMP_ISO8601:timestamp} %{NOTSPACE:elb} %{IP:clientip}:%{INT:clientport:int} (?:(%{IP:backendip}:?:%{INT:backendport:int})|-) %{NUMBER:request_processing_time:float} %{NUMBER:backend_processing_time:float} %{NUMBER:response_processing_time:float} %{INT:response:int} %{INT:backend_response:int} %{INT:received_bytes:int} %{INT:bytes:int} "%{ELB_REQUEST_LINE}"
+ELB_ACCESS_LOG %{TIMESTAMP_ISO8601:timestamp} %{NOTSPACE:elb} %{IP:clientip}:%{INT:clientport:int} (?:(%{IP:backendip}:?:%{INT:backendport:int})|-) %{NUMBER:request_processing_time:float} %{NUMBER:backend_processing_time:float} %{NUMBER:response_processing_time:float} %{INT:response:int} %{INT:backend_response:int} %{INT:received_bytes:int} %{INT:bytes:int} "%{ELB_REQUEST_LINE}" %{QS:agent} %{NOTSPACE:cipher} %{NOTSPACE:protocol}
```
jordansissel: Do all ELB access logs contain this?

jordansissel: This change also breaks backwards compatibility if someone was using ELB_ACCESS_LOG and has logs taht don't include the now-required agent, cipher, and protocol fields.

salexpdx: @jordansissel, Amazon only lists one format for their access logs and there doesn't appear to be any option to change back to the old format so I believe that all ELB access logs now contain this information.  

I can see the value in leaving ELB_ACCESS_LOG as is and creating a new format like ELB_ACCESS_LOG_EXTENDED_CIPHER that contains the new format.  Then users can decide to switch if they want the additional information included.

Let me know what you would prefer and I will update the pull request if needed.

joshuaspence:
+1 on this pull request.

kares:
believe these were addressed at https://github.com/logstash-plugins/logstash-patterns-core/pull/287, there's a spec where the new fields such as [tls](https://github.com/logstash-plugins/logstash-patterns-core/pull/287/files#diff-3dc1f8ed0fe0665d858649daae879afae6b4e0af80c2e55b13b1d425746821e0R116) are tested.
new captures will be available with an ECS compatible set of patterns. closing this since it would break compatibility.

PR #107, opened 2015-09-24
LOGSTASH-2198 added '_' to grok hostname pattern

mihaiplesa:


jordansissel:
Change looks ok, but I haven't tested it. There's no tests added in this PR either, can you add one to test this new case?

mihaiplesa:
Added test for verifying if underscore is now allowed in hostname, all tests pass.

purbon:
please jenkins , test this.

spec/patterns/core_spec.rb
```diff
@@ -4,7 +4,7 @@
 
 describe "SYSLOGLINE" do
 
-  let(:value)   { "Mar 16 00:01:25 evita postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]" }
+  let(:value)   { "Mar 16 00:01:25 UNKNOWN_HOSTNAME postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]" }
```
purbon: I like this addition, but can we please keep both? like this we're sure we match names with ```_``` and without.

purbon:
@mihaiplesa thanks a lot for your contribution, just added an small comment. As soon as this is ammended I'm good on merging this.

purbon:
please jenkins, test this.

mihaiplesa:
@purbon Done.

alibby:
Hostnames/ dns names beginning with an underscore character are also valid.  In fact, as odd as it sounds they can begin with a dash as well.

osek666:
What next with it?

vitalyrepin:
This pattern does not match the following hostname:

lb._dns-sd._udp.0.178.168.192.in-addr.arpa.

I suggest to change HOSTNAME pattern to the following value:

HOSTNAME \b(?:[0-9A-Za-z_][0-9A-Za-z_-]{0,62})(?:\.(?:[0-9A-Za-z_][0-9A-Za-z_-]{0,62}))*(\.?|\b)

wyardley:
There's a difference between hostnames and DNS records. Underscores are used in certain types of DNS records exactly _because_ they're not valid in hostnames. 

https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_host_names
https://tools.ietf.org/html/rfc952
https://tools.ietf.org/html/rfc1123

I would guess that for most of the things HOSTNAME is used in Logstash for (matching system hostnames and the RHS of email addresses), the stricter definition is better. It might be better to have a separate pattern (not named 'HOSTNAME'), if there's a need to match things that are valid DNS records.

jsvd:
closing due to @wyardley's argument: "hostname labels may contain only the ASCII letters 'a' through 'z' (in a case-insensitive manner), the digits '0' through '9', and the hyphen ('-')"
I'm sure open to a PR for DNS records, that we can gradually incorporate in "upstream" grok patterns

PR #109, opened 2015-10-02
Update and add java class and package patterns

### LINKED ISSUE #105: Errors in java patterns
The patterns
- JAVACLASS
- JAVAFILE
- JAVASTACKTRACEPART

in `patterns/java` are have multiple and non unique definitions. This causes issues for example as the second definition of JAVACLASS (which takes precedence) is defined as

```
JAVACLASS (?:[a-zA-Z0-9-]+\.)+[A-Za-z0-9$]+
```

is wrong. Class qualifiers should be optional and identifiers should be `[a-zA-Z$_][a-zA-Z$_0-9]*` as defined first time.

gustavkarlsson:
Added JAVARESERVEDWORD matching any reserved word in the language such as "null", "true" and "for"
Added JAVAIDENTIFIER matching a "section" in a FQCN and not accepting JAVARESERVEDWORD
Added JAVAPACKAGE matching a string of JAVAIDENTIFIERs joined by periods
Added JAVASIMPLECLASS matching the last section of a FQCN
Updated JAVACLASS to equal JAVAPACKAGE
Remove double definitions of JAVACLASS, JAVAFILE and JAVASTACKTRACEPART
Removed useless JAVALOGMESSAGE and replaced it with Oniguruma syntax

Fixes #105

### DIFF
--- a/patterns/java
+++ b/patterns/java
@@ -1,20 +1,19 @@
-JAVACLASS (?:[a-zA-Z$_][a-zA-Z$_0-9]*\.)*[a-zA-Z$_][a-zA-Z$_0-9]*
+JAVARESERVEDWORD null|true|false|abstract|continue|for|new|switch|assert|default|if|package|synchronized|boolean|do|goto|private|this|break|double|implements|protected|throw|byte|else|import|public|throws|case|enum|instanceof|return|transient|catch|extends|int|short|try|char|final|interface|static|void|class|finally|long|strictfp|volatile|const|float|native|super|while
+JAVAIDENTIFIER (?!%{JAVARESERVEDWORD})[a-zA-Z_$][a-zA-Z0-9_$]*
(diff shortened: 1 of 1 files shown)


gustavkarlsson:
I've signed the CLA now. Can I request the same commit to be pulled again without change somehow?

purbon:
please jenkins, test this.

purbon:
@gustavkarlsson thanks a lot for your contribution. As a compliment to your PR, it would be very nice of you if you can extend the current test for this patterns. Like this we'll be sure regressions are not introduced anymore. Makes sense? please let me know in case you need some assistance with it, welcome to provide guidance!

gustavkarlsson:
@purbon Makes total sense, but I'm afraid I'm no ruby programmer so I would probably just make a mess of it. If someone could get me started with a sample however, maybe I could do it. What I'm trying to say is that the guidance you're offering is welcome :)

PR #106, opened 2015-09-22
COMMONAPACHELOG: match email address in auth field

mattjbray:
#3 matches email addresses in the `ident` field.  Also allow email addresses in the `auth` field.

PR #115, opened 2015-10-29
Allowing for email in the auth field of combined Apache httpd log

oozie:
email as username in the auth field of the apache log is not accepted by grok patterns (`%{USER:auth}` corresponds to `[a-zA-Z0-9._-]+`) and logstash has a hard time parsing such entries. 

To reproduce:
* follow Apache logstash setup instructions https://www.elastic.co/guide/en/logstash/current/config-examples.html
* Set up basic authentication on your Apache HTTPD
* Try to authenticate with email address as username
* Navigate to kibana interface to find the entry. Its fields will be empty and it will be tagged as `_grokparsefailure`.


oozie:
This is a dupe of https://github.com/logstash-plugins/logstash-patterns-core/pull/106

Issue #112, opened 2015-10-22
Postgresql grok pattern fails


grahamrhay:
The postgres grok pattern (https://github.com/elastic/logstash/blob/v1.4.2/patterns/postgresql#L2) doesn't seem to be compatible with postgres 9.4. I'm not sure when the format changed.

Specifically, the pid at the end is no longer there. This is what I'm using:

%{DATESTAMP:timestamp} %{TZ} %{DATA:user_id} %{GREEDYDATA:connection_id} %{DATA:level}: %{GREEDYDATA:msg}

grahamrhay:
I can create a PR with that, but it's probably worth collating a few log messages from different versions of postgres first

purbon:
@grahamrhay thanks a lot for your contribution. Yeah! having a few log lines as examples would be awesome.

grahamrhay:
9.4:

2015-10-22 08:57:03 UTC [8850-3] foo@bar STATEMENT:  INSERT INTO ...
2015-10-22 08:57:03 UTC [8850-2] foo@bar DETAIL:  Key already exists.
2015-10-22 08:57:03 UTC [8850-1] foo@bar ERROR:  duplicate key value...

Issue #67, opened 2015-07-10
why is logstash parsing the year only as 2-digit?


csamsel:
I was using the following snippet to parse a customized size postgres logfile:
```
                grok {
                        match => {
                                "message" => [
                                  "%{DATESTAMP:timestamp_psql} %{TZ:tz} ...
```
which worked very well. As it turned out, sometimes postgres is using multiline, so my first shot was:

                multiline {
                        pattern => "^%{DATESTAMP}.*"
                        what => previous
                        negate => true
                }

which did not work. Looking at the JSON i found:

"timestamp_psql": "15-07-10 09:31:57.030 UTC",

so the leading 20 is discarded. I mean, for most logfiles this should be totally fine, but for me it was very confusing. I guess grok somehow ignores leading and trailing data for pattern matching.
Im now using 

                multiline {
                        pattern => "^20%{DATESTAMP}.*"
                        what => previous
                        negate => true
                }

as multiline filter (it works). but still thats wierd.

markwalkom:
Please join us in #logstash on Freenode or at https://discuss.elastic.co/ for troubleshooting help, we reserve Github for confirmed bugs and feature requests :)

csamsel:
Thanks for your response. Well, I figured it out by constructing the mentioned work around so i'm not really seeking for troubleshooting help.
I'm wondering if this is  designed intentionally this way, if not i'd consider it worth looking into (i'd call it a bug).

markwalkom:
Ok, apologies if I misunderstood! :)

purbon:
@csamsel This looks to me like a bug, I need to run a bit more rest, but a quick run of the YEAR expression at rubular.com matched for the 4 digits, so I would call it a bug for now.

Thanks a lot for your time and report!

purbon:
Hi @csamsel I was doing more test on your issue, specially with the DATESTAMP and seeing the output of the grok, all worked for me as expected. would you be able to provide me a sample log line? this would be super useful to actually validate if this is a grok error or a multiline one.

Thanks

csamsel:
Hi,
Yes this issue only rises with multiline parsing because in standard scenario grok will discard the leading 20 of date variables and the pattern will still match with 2-digit years. I was wondering if there is any reason for only groking and storing only 2-digit year numbers instead of the full year? btw this will als break in case old logfiles from 19XX are parsed.

[...]

PR #124, opened 2016-01-13
Adding DATE_8601 to the list of default Grok patterns

berglh:
On line 65 of this pattern file is the following comment:
```
 # datestamp is YYYY/MM/DD-HH:MM:SS.UUUU (or something like it)
```
It turns out this format in this comment isn't actually covered with YYYY MM DD with any other delimiter other than hyphens (such as in TIMESTAMP_ISO8601.

There are a lot of log files that contain `YYYY/MM/DD`, not just the official ISO8601 time stamp.

I made the `DATE_8601` pattern and suffixed it to the `DATESTAMP` pattern so when `DATESTAMP` is use, it covers all three common use formats. The `YYYY MM DD` is in used in China, Korea, Taiwan, Hungary, Iran, Japan, Lithuania due to `ISO 8601`. Also plenty of other countries use this format. I think for completeness of the basic date patterns it has a place here with all three common delimiters.

I think perhaps `DATE_8601` is the wrong name for this pattern. Using slashes or dot delimiters is not an official 8601 format according to [ISO8601](http://metric1.org/8601.pdf).

I'm suspecting we should change this, but I don't know what a suitable suffix would be as it's hard to aggregate the aforementioned countries into a region code like the other date patterns in the file.


berglh:
I did end up signing the CLA.

tedder:
This would address #67 and #112. I'm not sure if it resolves it, as .. wouldn't the `DATE_*` entries match before the third option is hit? The sane way to do it would be to add a "mandatory 4-digit year" entry to pick it off first:
```
YEAR4 (?>\d{4})
DATE_8061 %{YEAR4}[./-]%{MONTHNUM}[./-]%{MONTHDAY}
DATE %{DATE_8061}|%{DATE_US}|%{DATE_EU}
```

Issue #131, opened 2016-02-12
grok-patterns: firewalls, add SuSEfirewall2


dwbenjamin:
Add an additional pattern for SuSEfirewall2 format to provide more detailed searches.

PR #132, opened 2016-02-13
Added SuSEfirewall2 pattern

dwbenjamin:
#131


dwbenjamin:
I have now signed two of these and have received both confirmations:   Individual Contributor License Agreement under CCLA between elasticsearch BV and David Benjamin is Signed and Filed!

From: CLA Committers (elasticsearch BV)
To: CLA Committers and David Benjamin

karmi:
Hi @dwbenjamin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-patterns-core/pull/132.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

dwbenjamin:
I have added my email addresses as requested.

Thanks,
   David.

PR #133, opened 2016-02-20
Removed extra space in ASA-6-106015 rule

radu-gheorghe:
The %{CISCOFW106015} rule fails to parse a message like 'Deny TCP (no connection) from 192.168.150.65/2278 to 64.101.128.83/80 flags RST on interface inside' and it's because there are two spaces between the flags and "on interface". Removing that space makes it match (checked on http://grokdebug.herokuapp.com).

### DIFF
--- a/patterns/firewalls
+++ b/patterns/firewalls
@@ -36,7 +36,7 @@ CISCOFW106006_106007_106010 %{CISCO_ACTION:action} %{CISCO_DIRECTION:direction} 
 # ASA-3-106014
 CISCOFW106014 %{CISCO_ACTION:action} %{CISCO_DIRECTION:direction} %{WORD:protocol} src %{DATA:src_interface}:%{IP:src_ip}(\(%{DATA:src_fwuser}\))? dst %{DATA:dst_interface}:%{IP:dst_ip}(\(%{DATA:dst_fwuser}\))? \(type %{INT:icmp_type}, code %{INT:icmp_code}\)
 # ASA-6-106015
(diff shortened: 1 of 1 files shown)


ph:
@radu-gheorghe Would you mind creating a test for this changes to make sure we don't break anything in the future/past? If you need need help to get them running I can help you with that.

radu-gheorghe:
@ph sorry for the late reply. Sure, I'd gladly add a test, though I'm not 100% sure I'd do it right. It sounds like this file needs to be changed: https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/firewalls_spec.rb

Right?

And a new test would be added with a block like this?
```
  let(:pattern106015) { "CISCOFW106015" }  # assign the grok pattern to a name

  context "parsing a 106015 message" do   # rule type goes here

    let(:value) { "Deny TCP (no connection) from 192.168.150.65/2278 to 64.101.128.83/80 flags RST on interface inside" }  # this would be the original message

    subject     { grok_match(pattern106015, value) }  # specify the pattern name here, as defined before

    it { should include("interface" => "inside") }  # if the last field matches what we expect, it should be OK

    it "generates a message field" do
      expect(subject["message"]).to include("Deny TCP (no connection) from 192.168.150.65/2278 to 64.101.128.83/80 flags RST on interface inside")  # original message foes here
    end
  end
```

Am I getting it right?

jsvd:
@radu-gheorghe yep! that's it! sorry for the delay, but could you add that test so we can merge this? thanks!

radu-gheorghe:
Sure: I've added the test as https://github.com/logstash-plugins/logstash-patterns-core/pull/171

Thanks for looking into this!

jsvd:
may I ask you to close this PR and also add the pattern change in #171? thanks

radu-gheorghe:
Sure. Closing this one in favor of https://github.com/logstash-plugins/logstash-patterns-core/pull/171 then

PR #134, opened 2016-03-15
Add default nginx access log pattern

germaaan:


wyardley:
I was using the Apache Combined pattern, and seemed to mostly match. Which specific things are you changing (looks like you're using Nginx w/ normal combined log format)?

Not affiliated with elastic, but just curious, where is the NGUSER pattern coming from, since it doesn't seem to be in your changeset?

I was experiencing an issue due to email addresses in the authenticated username field, and submitted a pull request for that here.
https://github.com/logstash-plugins/logstash-patterns-core/pull/142

PR #135, opened 2016-03-17
Adding new patterns

Lowess:
@jordansissel 

This is my new branch to follow up on [1298](https://github.com/elastic/logstash/pull/1298). I am not working intensively on Logstash right now that's why it took me some time to update my PR. 

It would be a shame to not benefit from pre-made patterns because I think it's a great feature in Logstash.

Let me know if I am missing something.

Regards,
Florian


Lowess:
@karmi I did sign the CLA I do not know why it says I haven't... Any idea how to solve that?

karmi:
Hi @Lowess, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-patterns-core/pull/135.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #137, opened 2016-03-24
revert broken commit and fix pattern_path

colinsurprenant:
commit a273a051d8a0ef2e44ea9ceeb4d103902127ccc0 breaks most specs while trying to solve one specific spec. The commit is reverted and the broken spec disabled. 

Also fixes the missing `pattern_path` which a temporary fix to have the spec pass for an urgent mass-publish requirement. relates to logstash-plugins/logstash-filter-grok#72


suyograo:
LGTM

colinsurprenant:
pushed version 2.0.5

PR #139, opened 2016-04-01
Be able to parse timestamp where minute has only one digit.

samuelb:
I have a case where a application produces log entries where the timestamp has only one digit. Those entries can't be parsed.

```
2016-4-1 7:9:23 ...   <-- grokparsefailure
2016-4-1 7:13:42 ... 
```

This adjustment in the grok-patterns will allow to parse timestamps with one and two digit minute fields.


samuelb:
Well, on the other hand, ISO-8601 says that those time fields needs to be zero-padded.

Issue #141, opened 2016-04-07
user in COMMONAPACHELOG not matching email address


wyardley:
I am using nginx, but with NCSA combined format, so think it should match. I believe the 'auth' line in Apache can also be formatted as an email address.

See this example, for instance.
`10.0.0.1 - username@example.com [07/Apr/2016:18:42:24 +0000] "GET /bar/foo/users/1/username%40example.com/authenticate?token=blargh&client_id=15 HTTP/1.1" 400 75 "" "Mozilla/5.0 (iPad; CPU OS 9_3_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13E238 Safari/601.1"`

wyardley:
https://github.com/logstash-plugins/logstash-patterns-core/pull/142

magnusbaeck:
@andrewvc, I believe this issue can be closed.

PR #142, opened 2016-04-07
switch USER to HTTPDUSER for "auth" field (match email addresses)

wyardley:
See https://github.com/logstash-plugins/logstash-patterns-core/issues/141

This log fails to parse for me with the COMMONAPACHELOG / COMBINEDAPACHELOG grok rules:

```10.0.0.1 - username@example.com [07/Apr/2016:18:42:24 +0000] "GET /bar/foo/users/1/username%40example.com/authenticate?token=blargh&client_id=15 HTTP/1.1" 400 75 "" "Mozilla/5.0 (iPad; CPU OS 9_3_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13E238 Safari/601.1"```

### DIFF
--- a/patterns/grok-patterns
+++ b/patterns/grok-patterns
@@ -90,7 +90,7 @@ QS %{QUOTEDSTRING}
 
 # Log formats
 SYSLOGBASE %{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{SYSLOGPROG}:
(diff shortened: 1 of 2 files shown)


wyardley:
Signed CLA.

wyardley:
FWIW, I was able to generate an Apache 2.4 log entry with an email address in the username field as well
```10.0.0.1 - user@example.com [07/Apr/2016:12:49:36 -0700] "GET /authtest/ HTTP/1.1" 200 679```

wyardley:
I am not setup to develop on Logstash, so might be some minor issues (wasn't able to test), but added a basic test for this as well.

Issue #159, opened 2016-06-24
%UNIXPATH can cause RegEx DOS


jsvd:
moved from https://github.com/elastic/logstash/issues/5458

Repro is very simple.   This will peg your CPU:

```
input {
        stdin {
                add_field => { "test" => "/web/asdf/my_input_file_000223.xml name"  }
        }
}


filter {
        grok {
                match => {
                        "test" => [
                                "%{UNIXPATH}/asdf/my_%{DATA}_file_%{DATA:fileName} name"
                        ]
                }
        }
}

output {
        stdout {
          codec  => rubydebug {
            metadata => true
          }
        }
}
```

The more `/asdf/asdf` you add, the longer it takes.

It works without delay on this [online grok debugger](http://grokdebug.herokuapp.com/) app:

![grokdebug](https://cloud.githubusercontent.com/assets/4387023/15861888/c8a1dc04-2c9b-11e6-83e7-9630d8a33bd1.png)

Investigating, it was found that the UNIXPATH is different between tools:

Online App: `UNIXPATH (?>/(?>[\w_%!$@:.,-]+|\\.)*)+` (Works Better)
Logstash: `UNIXPATH (/([\w_%!$@:.,~-]+|\\.)*)+` (Can Blow Up)

purbon:
This happen because the current UNIXPATH expression is getting "confused" by actual input data, with:

expression: ` UNIXPATH (/([\w_%!$@:.,+~-]+|\\.)*)+ `
input data: ` /web/asdf/my_input_file_000223.xml name `
grok: ` %{UNIXPATH}/asdf/my_%{DATA}_file_%{DATA:fileName} name `

the regexp is trying to match as far as it can, non understanding the first '/' as a new non related token as is the intention for it. We can see this is true by doing a simply change:

input data: ` "/web /asdf/my_input_file_000223.xml name"` 
grok: `%{UNIXPATH} /asdf/my_%{DATA}_file_%{DATA:fileName} name`

so the current expression knows when it has to finish consuming characters.

purbon:
interesting fact, from the current pattern rspec tests:


1) will generate a timeout error, as with the previous example.

```
 context "when using recursive paths" do

    let(:pattern) { "%{UNIXPATH}/bar/%{DATA:fileName}" }
    let(:value)   { "/foo/bar/my_input_file_000223.xml" }

    it "should match the path expression" do
      expect(grok_complex_match(pattern,value)).to pass
    end
  end
[...]

PR #166, opened 2016-08-03
Fix for unixpath endless backtracking problem.

purbon:
Fixes bug #159 related to unix path backtracking until timeout due to not having proper finish condition. This commit add an atomic grouping for the unix path expression, so the path is required to match see http://ruby-doc.org/core-2.3.0/Regexp.html#class-Regexp-label-Atomic+Grouping
for more details.

Details around the issue could be found at #159.

### DIFF
--- a/patterns/grok-patterns
+++ b/patterns/grok-patterns
@@ -35,7 +35,7 @@ HOSTPORT %{IPORHOST}:%{POSINT}
 
 # paths
 PATH (?:%{UNIXPATH}|%{WINPATH})
-UNIXPATH (/([\w_%!$@:.,+~-]+|\\.)*)+
+UNIXPATH (?>\/[[\w_%!$@:.,-]+|\\.]*)+
 TTY (?:/dev/(pts|tty([pq])?)(\w+)?/?(?:[0-9]+))
 WINPATH (?>[A-Za-z]+:|\\)(?:\\[^\\?*]*)+
 URIPROTO [A-Za-z]+(\+[A-Za-z+]+)?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 4.0.3
+  - Fix bug https://github.com/logstash-plugins/logstash-patterns-core/issues/159 with the introduction of atomic grouping (http://ruby-doc.org/core-2.3.0/Regexp.html#class-Regexp-label-Atomic+Grouping) for the unix path expression, this makes hte engine not to backtrack until timeout without specific reason, so expression gets optimized to find exit condition in such situations.
+
 ## 4.0.2
   - Relax constraint on logstash-core-plugin-api to >= 1.60 <= 2.99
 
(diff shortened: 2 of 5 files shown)


purbon:
@jordansissel @suyograo can you guys provide some feedback here? would be important to get this fix for release.

patterns/grok-patterns
```diff
@@ -35,7 +35,7 @@ HOSTPORT %{IPORHOST}:%{POSINT}
 
 # paths
 PATH (?:%{UNIXPATH}|%{WINPATH})
-UNIXPATH (/([\w_%!$@:.,+~-]+|\\.)*)+
+UNIXPATH (?>\/[[\w_%!$@:.,-]+|\\.]*)+
```
jordansissel: I'm not sure I'm reading this correctly. Can you help me understand?

I see two main changes. 

One, using an atomic group. I like. 

Two, it seems like the character class `[....]` was wrapped with another `[...]` and includes the `+|\\.` bits, which feels incorrect to me.

Here's a difference of before/after. It looks like the "after" is wrong. It shouldn't be one big charater class. Should it?

```diff
-[\w_%!$@:.,+~-]+|\\.
+[[\w_%!$@:.,-]+|\\.]
```

purbon: Will try to explain what I did here:

> Two, it seems like the character class [....] was wrapped with another [...] and includes the +|\\. bits, which feels incorrect to me.

I changed the first level parenthesis group in favor of the brackets to there is only one "bigger" group, but looks in the process of cleaning this up I eat the `+~` characters that I understand are meaningful for unix file patterns.

Keeping the changes to:

* The atomic group.
* Only having one group.

Makes sense for you?

jordansissel: @purbon I can explain more. In regexp, the brackets `[]` and parentheses `()` mean completely different things. Parentheses are syntax for grouping. Brackets are syntax for character classes.

These two things mean completely different things: `[abcd]` vs `(abcd)`. The first (brackets) matches a single character that can be either a, b, c, or d. The second (parentheses) matches exactly the text `abcd`.

So my concern about the `()` to `[]` change you made:

```diff
-[\w_%!$@:.,+~-]+|\\.
+[[\w_%!$@:.,-]+|\\.]
```

The way I read this regexp, the former (removed code) means basically `one or more of the following characters: word character, `_`, `%`, .... `-`` _OR_ a literal `.` character.

The second (new code) means basically: `match one single character that can be [some of those characters]`

Example:

```
# Old code, matching a single part of the path.
>> /[\w_%!$@:.,+~-]+|\\./.match("/hello/world")
=> #<MatchData "hello">

# New code,
>> /[[\w_%!$@:.,-]+|\\.]/.match("/hello/world")
=> #<MatchData "h">
```

The character class is changed in this PR by adding, i think, left bracket, pipe, and backslash.

This PR description looks like it is supposed to fix the behavioral bug (ReDOS), not add new things that can match. Right?

kares:
closing in favour of https://github.com/logstash-plugins/logstash-patterns-core/pull/292

PR #167, opened 2016-08-09
add pattern for cloudfront access log

bdashrad:
This PR adds a pattern for AWS CloudFront access logs.

Some of the common patterns I found online were incorrectly parsing the date field.


jsvd:
LGTM

PR #169, opened 2016-08-17
Added Redis Monitor Log format

rijnhard:

A common pattern is to pipe redis monitor output to a logfile for auditing or debugging purposes. 
The format of these logs though is a bit different, and to make it more confusing the command logged impacts the number of parameters. 
So as a simpler way to handle this I extract as much info as possible without getting into the intricacies of the the commands themselves: 

This was the data i tested against (with some manufactured pieces to cater for unknown future variations)

```
1470637867.953466 [0 195.168.1.1:52500] "info"
1339518083.107412 [0 127.0.0.1:60866] "keys" "*"
1470637875.777457 [15 195.168.1.1:52500] "select" "16"
1470637925.186681 [0 127.0.0.1:39404] "rpush" "my:special:key" "{\"data\":\"cdr\",\"payload\":\"json\"}"
1470637875.777457 [15 195.168.1.1:52500] "intentionally" "broken" "variadic" "log" "entry"
```

The params are simply not parsed, the more I tried to parse it the more it became apparent that you would have to handle it specifically per command, or at least for categories of commands.

### DIFF
--- a/patterns/redis
+++ b/patterns/redis
@@ -1,3 +1,3 @@
 REDISTIMESTAMP %{MONTHDAY} %{MONTH} %{TIME}
 REDISLOG \[%{POSINT:pid}\] %{REDISTIMESTAMP:timestamp} \* 
-
+REDISMONLOG %{NUMBER:timestamp} \[%{INT:database} %{IP:client}:%{NUMBER:port}\] "%{WORD:command}"\s?%{GREEDYDATA:params}
--- a/spec/patterns/redis_spec.rb
+++ b/spec/patterns/redis_spec.rb
@@ -1,0 +1,171 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/patterns/core"
+
+describe "REDISTIMESTAMP" do
+
+  let(:value) { '14 Nov 07:01:22.119'}
+  let(:pattern) { "REDISTIMESTAMP" }
+
+  it "a pattern pass the grok expression" do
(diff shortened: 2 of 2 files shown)


rijnhard:
I have signed the CLA

rijnhard:
I haven't written tests as none exist for redis as it stands, it will be trivial to do so if required.

jsvd:
Would you mind creating one test for this? I'll merge it asap

rijnhard:
@jsvd will do

karmi:
Hi @rijnhard, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-patterns-core/pull/169.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rijnhard:
@jsvd tests added to the best of my ability (my ruby isn't great)
@karmi done. 

also side note: running the tests is really easy with docker 

```bash
> docker run -ti -v "$PWD":/usr/src/logstash -w /usr/src/logstash jruby:9 bash
root@a47a4777a5d9:/usr/src/logstash> bundle install 
root@a47a4777a5d9:/usr/src/logstash> bundle exec rspec 
```

You can collapse it into one command but at how long the install takes I would not reccomend

jsvd:
wow great work @rijnhard, thanks for the effort of writing all these tests

PR #174, opened 2016-09-23
Httpd types

chernjie:
Dependent on #102, breaking compatibility by changing types on the fields


kares:
we've addressed type-casting for the updated ECS `HTTPD` pattern set at https://github.com/logstash-plugins/logstash-patterns-core/pull/267
for backwards compatibility the legacy set will keep on NOT converting these captures. 
thanks for the effort though.

PR #177, opened 2016-10-03
Add grok patterns to parse Barracuda Spam Firewall Mail and Web Syslog

anhlqn:
The patterns should work properly on BSF firmware v7.x.x.xxx. No failed grok in tags so far.

Each field in Mail syslog contains either a value or "-", so I use NOTSPACE pattern. I'm not sure if NOTSPACE affects grok performance compared to something like a custom pattern like INT|- to parse either an integer or a "-".

### DIFF
--- a/patterns/bsf
+++ b/patterns/bsf
@@ -1,5 +1,4 @@
 # Barracuda Spam Firewall Grok Patterns
-
 # Fields: action_id, client_ip, client_name, destination, delivery_detail, encryption, message_id, process, reason_extra, reason_id, recipient, score, sender, service, message_size, subject, unix_start_time, unix_end_time
 
 BSF_SERVICE RECV|SCAN|SEND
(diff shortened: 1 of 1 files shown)


karmi:
Hi @anhlqn, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-patterns-core/pull/177.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

anhlqn:
Can I close this pull request and create another under the email in my Github profile?

anhlqn:
I hope the email has been updated. I'm pretty new to Github.

anhlqn:
Is it possible to completely remove this pull request? My work email should not be tied to my personal github activities. Thanks.

PR #179, opened 2016-10-12
Fix grokparsefailure due to truncated http_request.

ashemedai:
The HAProxy documentation lists for httplog's http_request:

"This field might be truncated if the request is huge and does not fit
in the standard syslog buffer (1024 characters). This is the reason why
this field must always remain the last one."

If the field is truncated the ending quote will never be present and
thus fail the preceding regular expressions. By making the ending quote
optional we fill the http_request field, and http_version will be null.


ashemedai:
CLA is currently processing.

ashemedai:
@jsvd Haven't looked at the tests part yet. But I ran this without any problems against a production 1.5.14 instance's logfile and now have no more grokparsefailures that I noticed.

ashemedai:
Took me a while, but I added a test case.

ashemedai:
@jsvd The Travis CI build setup seems broken for two of the three environments:

jruby: warning: unknown property jruby.cext.enabled
bundler: failed to load command: rspec (/home/travis/build/logstash-plugins/logstash-patterns-core/vendor/bundle/jruby/2.3.0/bin/rspec)

The one that works confirms the PR is not breaking anything.

jsvd:
@ashemedai thank you for the contribution and effort :)

ashemedai:
@jsvd Thanks for merging, now I can get rid of my local hacks. ;)

Someone needs to fix those two Travis CI instances though. :\

PR #180, opened 2016-10-30
Adding VPC flowlogs to AWS patterns

Bizzelicious:
Added pattern for VPC flowlogs in AWS.


jsvd:
Please add a test so we can merge this: the spec suite has many examples like https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/firewalls_spec.rb#L5-L21

PR #181, opened 2016-10-30
Added openvpn patterns

Bizzelicious:
Added patterns for Open VPN.


jsvd:
@Bizzelicious thanks for the contribution! can you add a few specs to show what kind of lines these patterns match? there are a lot of simple examples in the `spec` directory. essentially all you need is:

```ruby
describe "PATTERN" do
  it "matches an open vpn log message" do
    expect(subject).to match("example of open vpn log message")
  end
end
```

PR #184, opened 2016-11-14
According to rcf5424 IP address should be accepted

flysen:
https://www.rfc-editor.org/rfc/rfc5424.txt 

- **6.2.4.  HOSTNAME** 
IP address should be accepted not only HOSTNAME


jsvd:
@flysen thanks for the contribution! can I ask you to just add a spec so I can merge it?

[Here](https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/syslog_spec.rb)'s an example:

```
 it "matches prog from ansible" do
    expect(subject).to match("May 11 15:40:51 meow.soy.se ansible-<stdin>: Invoked with filter=* fact_path=/etc/ansible/facts.d")
  end
```

flysen:
@jsvd I’m not sure how to do that. Doesn't it need a new description SYSLOG5424LINE? 
I can provide log lines for tests.

jsvd:
correct, all you need to add in that file is, inside the main describe block, another describe for this particular pattern:

```ruby
describe "SYSLOGLINE" do  
  describe "SYSLOG5424BASE" do
     it "matches host names in the syslog base pattern" do
       expect(subject).to match("log line here")
    end
  end
end
```

jsvd:
interesting, while reviewing this PR I noticed that the HOSTNAME pattern actually accepts any sequence of numbers with dots in them, which means it accepts ipv4 addresses (but not ipv6).

example:

```
describe "HOSTNAME" do

  it "matches a simple hostname" do
    expect(subject).to match("google.pt")
  end

  it "matches an ipv4 address" do
    expect(subject).to match("213.13.11.41")
  end

  it "matches a sequence of numbers separated by dots" do
    expect(subject).to match("213.13.11.41.12.123.5.1.1.5.1")
  end
end
```

PR #189, opened 2016-12-06
Minor improvement to S3 grok patterns

joshuaspence:
Allows for the `requester` and `key` fields to be `-`, which is equivalent to `null`.


kares:
a new pattern set with captures revisited (to be ECS compliant) had been [merged](https://github.com/logstash-plugins/logstash-patterns-core/pull/287).
the optional fields mentioned here are [handed as expect](https://github.com/logstash-plugins/logstash-patterns-core/pull/287/files#diff-91e3fdee91aabacc758639784325d5b8072c4b699c1e92d68a4c2d786b268cc2R3). 
we're likely going to leave the legacy set as is (some users might assume the matching `-` behaviour as a feature).

thanks for your efforts though!

PR #195, opened 2017-02-03
Correct CISCOFW106023 pattern to use IP pattern instead of DATA

jjm:
When using firewall pattern `CISCOFW106023` as follows:

```ruby
    if "%ASA-" in [app_name] {
      mutate {
        add_tag => [ "Firewall", "ASA" ]
      }
      grok {
        match => {
          "message" => [
            "%{CISCOFW106023}"
          ]
        }
      }
    }
```

The matching of the `src_ip` and `dst_ip` fields using `DATA` instead of `IP` leads to _text_ in these fields when processing logs from Cisco ASA which have `names` command configured. This command replaces IP addresses in syslog messages with text. This does not happen with any other firewall  patterns as they all use `IP` instead.

This leads to errors when inserting these documents into elasticsearch when the `src_ip` and `dst_ip` fields are mapped to IP address, we are seeing these errors about ~8000 a day and elastic support has requested we fix our logstash configuration.


jjm:
Looks like the build is failing for all patterns and not just the one I changed. Likely a deeper issue here.

jjm:
So now failing with the following:

```
1) FIREWALLS parsing a 106023 message with a hostname should break the message up into fields
     Failure/Error: expect(subject['action']).to eq('Deny')
     
       expected: "Deny"
            got: nil
     
       (compared using ==)
     # ./spec/patterns/firewalls_spec.rb:127:in `block in (root)'
     # /home/travis/.rvm/gems/jruby-9.1.13.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block in (root)
```

So the unit tests checks for an ASA using `names` and the `dst_ip` contains text, just pushed a change to make the fields optional and update the unit tests. However it's likely a breaking change for some configurations, however if `src_ip` or `dst_ip` are an ip type in elasticsearch then the data will not get inserted anyway.

jjm:
I've changed jobs and no longer have access to the source repo (or access to any Cisco ASA's / using Logstash), so I can't address any comments on this PR if I received them. So I'll just close this off.

PR #197, opened 2017-02-24
Update URIPROTO to match RFC3986

olafz:
According to RFC 3986 (paragraph 3.1):
> Scheme names consist of a sequence of characters beginning with a letter and followed by any combination of letters, digits, plus ("+"), period ("."), or hyphen ("-").

What is called a scheme in the RFC, is referred to as `URIPROTO` in Logstash patterns. For instance, a valid URI would be "android-app://...", but is currently not captured by the Grok pattern for URI. This PR fixes that.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -36,6 +36,7 @@ Contributors:
 * Matthew Baxa (mbaxa)
 * MikeSchuette
 * Nick Padilla (NickPadilla)
+* Olaf van Zandwijk (olafz)
 * Oluf Lorenzen (Finkregh)
 * Paul Myjavec (pmyjavec)
 * Pete Fritchman (fetep)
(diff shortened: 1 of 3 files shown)


olafz:
Hmmm; looks like the Travic CI build failed on a lot of patterns and not the one I changed...

jordansissel:
Can you add tests for this?

olafz:
@jordansissel I'm happy to add tests for this particular change, but that would then also be the only tests for `URIPROTO` or am I missing something?

jordansissel:
@olafz you are right, these would be the only tests for URIPROTO. I am OK with this PR only adding specifically URI schemes that are intended to be matched by changes in this PR.

olafz:
I added tests for this change, but the build fails. To verify that it wasn't my change, I ran tests on the current master branch, but they also fail with me. That makes it difficult to add tests :)

olafz:
All done

jordansissel:
LGTM

jordansissel:
@olafz thank you for helping improve Logstash!

PR #199, opened 2017-04-27
HTTPD24_ERRORLOG without tid and errorcode

mu5h3r:
Error message example: [Thu Apr 27 10:39:46.719636 2017] [php7:notice] [pid 17] [client 10.255.0.3:49580] Test error log record


jordansissel:
LGTM

PR #201, opened 2017-05-11
Add patterns for some message IDs

anhlqn:
- Add patterns for some message IDs
- Add a pattern for message IDs and assign the value to field `message_id`. Having `message_id` field  allows us to group events together, and we can also use dictionary filter plugin to add Cisco [recommended action](http://www.cisco.com/c/en/us/td/docs/security/asa/syslog-guide/syslogs.html) based on this field.

If possible, I also want to update existing Cisco patterns to extract this `message_id` field. Cisco syslog is quite helpful for me to troubleshoot network issues.

### DIFF
--- a/patterns/firewalls
+++ b/patterns/firewalls
@@ -84,6 +84,8 @@ CISCOFW713172 Group = %{GREEDYDATA:group}, IP = %{IP:src_ip}, Automatic NAT Dete
 CISCOFW713202 %{CISCO_MESSAGE_ID:message_id}: IP = %{IP:src_ip}, %{GREEDYDATA:reason}. %{CISCO_ACTION:action} packet.
 # ASA-4-733100
 CISCOFW733100 \[\s*%{DATA:drop_type}\s*\] drop %{DATA:drop_rate_id} exceeded. Current burst rate is %{INT:drop_rate_current_burst} per second, max configured rate is %{INT:drop_rate_max_burst}; Current average rate is %{INT:drop_rate_current_avg} per second, max configured rate is %{INT:drop_rate_max_avg}; Cumulative total count is %{INT:drop_total_count}
+# ASA-5-750006
+CISCOFW750006 %{CISCO_MESSAGE_ID:message_id}: Local:%{IP:dst_ip}:%{INT:dst_port} Remote:%{IP:src_ip}:%{INT:src_port} Username:%{USERNAME:username} %{DATA}. Reason: %{CISCO_REASON:reason}
 #== End Cisco ASA ==
 
 # Shorewall firewall logs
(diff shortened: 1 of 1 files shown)


jordansissel:
It is difficult for me to evaluate this change because there are no tests available to verify it. If you added tests for the new behavior, I could more easily evaluate the change.

anhlqn:
What kind of test can I provide? Would some sample logs be good enough? The patterns have been tested on https://grokdebug.herokuapp.com/ and live system. On live system, I monitor the message ID and make sure there's no _grokparsefailure tag.

jsvd:
@anhlqn yes, a few log line examples would make it more explicit on what kind of lines these patterns would match. This serves both the purpose of confirming the lines match but also provide documentation. there are many examples in the spec files you can reuse, like https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/firewalls_spec.rb#L84-L101

anhlqn:
These message IDs follow only one format, so I think one log example should be enough.

CISCOFW713202
```
%{CISCO_MESSAGE_ID:message_id}: IP = %{IP:src_ip}, %{GREEDYDATA:reason}. %{CISCO_ACTION:action} packet.

%ASA-5-713202: IP = 1.2.3.4, Duplicate first packet detected.  Ignoring packet.
```

CISCOFW750006
```
%{CISCO_MESSAGE_ID:message_id}: Local:%{IP:dst_ip}:%{INT:dst_port} Remote:%{IP:src_ip}:%{INT:src_port} Username:%{USERNAME:username} %{DATA}. Reason: %{CISCO_REASON:reason}

%ASA-5-750006: Local:1.2.3.4:1234 Remote:5.6.7.8:5678 Username:5.6.7.8 IKEv2 SA UP. Reason: New Connection Established
```

CISCOFW402116
```
%{WORD:protocol}: Received an %{WORD:orig_protocol} packet \(SPI= %{DATA:spi}, sequence number= %{DATA:seq_num}\) from %{IP:src_ip} \(user= %{USERNAME:username}\) to %{IP:dst_ip}

%ASA-4-402116: IPSEC: Received an ESP packet (SPI= 0x11XX2233, sequence number= 0x11XX) from 1.2.3.4 (user= 1.2.3.4) to 10.0.0.1.  The decapsulated inner packet doesn't match the negotiated policy in the SA.  The packet specifies its destination as 2.3.4.5, its source as 1.2.3.4, and its protocol as icmp.  The SA specifies its local proxy as 2.3.4.5/255.255.255.255/ip/0 and its remote_proxy as 3.4.5.6/255.255.255.255/ip/0.
```

CISCOFW113009
```
%{CISCO_MESSAGE_ID:message_id}: %{CISCO_REASON:reason} \(%{WORD:src_name}\) for user = %{USERNAME:username}

%ASA-6-113009: AAA retrieved default group policy (client_name) for user = 1.2.3.4
```

CISCOFW605005
```
%{CISCO_MESSAGE_ID:message_id}: %{CISCO_REASON:reason} from %{IP:src_ip}\/%{INT:src_port} to %{DATA:dst_interface}:%{IP:dst_ip}\/%{DATA:service} for user "%{USERNAME:username}"

%ASA-6-605005: Login permitted from 1.2.3.4/12345 to some_interface:5.6.7.8/http for user "some_user"
```

PR #211, opened 2017-10-12
Type-cast various integer quantities to int.

MattSANU:
But don't type-cast data parsed using numeric patterns which aren't actually numbers, such as HTTP versions.


MattSANU:
The current build failures don't *appear* to be the result of these changes. But I confess I have no idea what *is* causing them.

jsvd:
First of all thanks for your contribution.
I'm a bit uncertain if this should be merged even though it makes sense most of the time. Pretty much all patterns on the repo don't perform casting, this would be a huge breaking change, and we don't see a lot of demand for this.

I'm tempted to wait for some feedback from the community, wdyt?

stevecrozz:
We're starting to index our logs using an elk cluster, and I'm noticing a lot of fields that are numeric but indexed as strings. This is not ideal as I may want to do math on things like HAProxy response times. I'm surprised at the lack of interest/support in this ticket. It makes me wonder if a large number of people are relying on https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic-field-mapping.html#numeric-detection.

The documentation here says "usually the correct solution is to map these fields explicitly" and I like the idea behind this PR because it does exactly that.

kares:
believe most (likely all) of these will be addressed in the effort to make captures ECS compliant: https://github.com/logstash-plugins/logstash-patterns-core/issues/278
the legacy patterns will match as before (wout type-casting) for compatibility, users looking for an improved experience are expected to opt in to our ECS counterparts.

PR #226, opened 2018-04-17
Add Cisco ASA messages to firewalls patterns.

fchouina:
Fix issue with ASA-3-313004.
Add Cisco ASA message patterns to firewalls patterns file:
- ASA-2-106017
- ASA-1-106101
- ASA-6-303002
- ASA-3-303006
- ASA-5-305013
- ASA-5-713257
- ASA-4-713903
- ASA-6-725001
- ASA-6-725002
- ASA-6-725007
- ASA-6-725016
- ASA-3-746016

### DIFF
--- a/patterns/firewalls
+++ b/patterns/firewalls
@@ -11,6 +11,8 @@ CISCO_REASON Duplicate TCP SYN|Failed to locate egress interface|Invalid transpo
 CISCO_DIRECTION Inbound|inbound|Outbound|outbound
 CISCO_INTERVAL first hit|%{INT}-second interval
 CISCO_XLATE_TYPE static|dynamic
+CISCO_XLATE_CATEGORY outbound static|identity|portmap|regular
+
 # ASA-1-104001
 CISCOFW104001 \((?:Primary|Secondary)\) Switching to ACTIVE - %{GREEDYDATA:switch_reason}
 # ASA-1-104002
@@ -37,6 +39,8 @@ CISCOFW106006_106007_106010 %{CISCO_ACTION:action} %{CISCO_DIRECTION:direction} 
 CISCOFW106014 %{CISCO_ACTION:action} %{CISCO_DIRECTION:direction} %{WORD:protocol} src %{DATA:src_interface}:%{IP:src_ip}(\(%{DATA:src_fwuser}\))? dst %{DATA:dst_interface}:%{IP:dst_ip}(\(%{DATA:dst_fwuser}\))? \(type %{INT:icmp_type}, code %{INT:icmp_code}\)
 # ASA-6-106015
 CISCOFW106015 %{CISCO_ACTION:action} %{WORD:protocol} \(%{DATA:policy_id}\) from %{IP:src_ip}/%{INT:src_port} to %{IP:dst_ip}/%{INT:dst_port} flags %{DATA:tcp_flags} on interface %{GREEDYDATA:interface}
+# ASA-2-106017
+CISCOFW106017 Deny IP due to Land Attack from %{IP:src_ip} to %{IP:dst_ip}
 # ASA-1-106021
 CISCOFW106021 %{CISCO_ACTION:action} %{WORD:protocol} reverse path check from %{IP:src_ip} to %{IP:dst_ip} on interface %{GREEDYDATA:interface}
 # ASA-4-106023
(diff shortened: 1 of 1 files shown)


fchouina:
I have now signed the CLA.

patterns/firewalls
```diff
@@ -55,10 +61,18 @@ CISCOFW302010 %{INT:connection_count} in use, %{INT:connection_count_max} most u
...
 # ASA-6-302020, ASA-6-302021
 CISCOFW302020_302021 %{CISCO_ACTION:action}(?: %{CISCO_DIRECTION:direction})? %{WORD:protocol} connection for faddr %{IP:dst_ip}/%{INT:icmp_seq_num}(?:\(%{DATA:fwuser}\))? gaddr %{IP:src_xlated_ip}/%{INT:icmp_code_xlated} laddr %{IP:src_ip}/%{INT:icmp_code}( \(%{DATA:user}\))?
+# ASA-6-303002
+CISCOFW303002 FTP connection from %{DATA:src_interface}:%{IP:src_ip}/%{INT:src_port} to %{DATA:dst_interface}:%{IP:dst_ip}/%{INT:dst_port}, user %{DATA:dst_user} %{DATA:ftp_action} file %{DATA:filename}
+# ASA-3-305006
+CISCOFW305006 %{CISCO_XLATE_CATEGORY:xlate_category} translation creation failed for %{WORD:protocol} src %{DATA:src_interface}:%{IP:src_ip}(?:/%{INT:src_port})?(?: %{DATA:src_fwuser})? dst %{DATA:dst_interface}:%{IP:dst_ip}(?:/%{INT:dst_port})?(?: %{DATA:dst_fwuser})? ?%{GREEDYDATA:additional_data}?
 # ASA-6-305011
 CISCOFW305011 %{CISCO_ACTION:action} %{CISCO_XLATE_TYPE:xlate_type} %{WORD:protocol} translation from %{DATA:src_interface}:%{IP:src_ip}(/%{INT:src_port})?(\(%{DATA:src_fwuser}\))? to %{DATA:src_xlated_interface}:%{IP:src_xlated_ip}/%{DATA:src_xlated_port}
-# ASA-3-313001, ASA-3-313004, ASA-3-313008
```
jsvd: I'm unfamiliar with the asa firewalls, but I'd like to avoid removing patterns as that can create troubles for folks already using this CISCOFW313001_313004_313008 pattern.

fchouina:
We could leave the name as is, but that pattern does not match ASA-3-313004 properly.
That is why I had to add a pattern specific for ASA-3-313004 right after.
Personally, I do not mind, but it does not document properly what is going on in the patterns matching if we leave it as is.

patterns/firewalls
```diff
@@ -79,8 +93,22 @@ CISCOFW602303_602304 %{WORD:protocol}: An %{CISCO_DIRECTION:direction} %{GREEDYD
 CISCOFW710001_710002_710003_710005_710006 %{WORD:protocol} (?:request|access) %{CISCO_ACTION:action} from %{IP:src_ip}/%{INT:src_port} to %{DATA:dst_interface}:%{IP:dst_ip}/%{INT:dst_port}
 # ASA-6-713172
 CISCOFW713172 Group = %{GREEDYDATA:group}, IP = %{IP:src_ip}, Automatic NAT Detection Status:\s+Remote end\s*%{DATA:is_remote_natted}\s*behind a NAT device\s+This\s+end\s*%{DATA:is_local_natted}\s*behind a NAT device
+# ASA-5-713257
+CISCOFW713257 Phase %{INT:phase} failure:  Mismatched attribute types for class %{DATA:class}:  Rcv'd: %{DATA:received_attribute}  Cfg'd: %{GREEDYDATA:configured_attribute}
```
jsvd: This pattern has a few double spaces, is this intended?

fchouina: Yes.
Here is a sample log message:
```
Phase 1 failure:  Mismatched attribute types for class Group Description:  Rcv'd: Group 2  Cfg'd: Group 5
```

jsvd: I see, my question came from doing a quick search online and finding references to that log entry with only 1 space: https://supportforums.cisco.com/t5/remote-access/ipsec-vpn-to-asa-5520/td-p/1528993

Maybe we can be safe here and replace these potential "more than one" spaces with [`%{SPACE}`](https://github.com/logstash-plugins/logstash-patterns-core/blob/master/patterns/grok-patterns#L15) ?

fchouina: We could use '\s+' (1 or more spaces) but not %{SPACE} because it is defined as '\s*' (zero or more spaces).
Using %{SPACE} could break the pattern.

fchouina: BTW, if it sheds more light on the subject, all the ASA-5-713257 messages I have ever seen (I have been parsing Cisco ASA logs for 3 years now) have always had the 2 spaces in those places.

colinsurprenant:
@fchouina  to continue the discussion on the `CISCOFW313001_313004_313008` pattern you suggest to remove: does that pattern causes any problem or conflict with the patterns you suggest adding? As @jsvd  mentioned, I would be cautious to not create BWC issue for anyone potentially using it. If you believe this is still a valid fix but somewhat unrelated to your proposed additions here, please submit a new PR just for that. Does that make sense?

fchouina:
No there is no conflict with the new ASA-3-313004 pattern. The original CISCOFW313001_313004_313008 pattern simply does not match any ASA-3-313004 logs.

colinsurprenant:
Ok @fchouina  thanks for confirming. So for the sake of keeping separate concerns into separate issues, could you please revert the `CISCOFW313001_313004_313008` pattern and create a new issue for it? Thanks so much.

joedissmeyer:
This is great! Thanks for contributing the additional filters @fchouina!

I have several more to contribute myself for these message patterns:
ASA-5-713050
ASA-5-713259
ASA-4-113019
ASA-5-713119
ASA-5-713120
ASA-5-713049
ASA-5-713076
ASA-5-713130
ASA-4-722012
ASA-4-722037
ASA-4-722041
ASA-4-722051
ASA-5-722010
ASA-5-722028
ASA-5-722032
ASA-5-722033
ASA-5-722034
ASA-5-737003
ASA-5-737034

Maybe I'll wait for this PR to be merged into master until I contribute? I don't want to submit another PR and cause anything to get lost.

Issue #227, opened 2018-04-23
SuSE Firewall 2, does not return the date / time of the log entry.


dwbenjamin:
logstash-patterns-core-4.1.2

SFW2 should return log entry date.

fixes https://github.com/logstash-plugins/logstash-patterns-core/pull/228

kares:
Thanks for the report David, this is expected to be resolved at https://github.com/logstash-plugins/logstash-patterns-core/pull/293

PR #228, opened 2018-04-23
SuSE Firewall 2, does not return the date / time of the log entry. #227

dwbenjamin:


jsvd:
CLA seems to be fine now, can you rebase the PR against master?

Also, can you add a sample log message in a test like we do for other patterns like [this one](https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/redis_spec.rb#L65-L73)?

Thanks!

dwbenjamin:
Sorry, I'm new to all this fancy "git stuff."

I added the test and trying again.
   DB

PR #229, opened 2018-05-08
SFW2

dwbenjamin:
SuSE Firewall 2, does not return the date / time of the log entry. #227


kares:
an alternative [approach](https://github.com/logstash-plugins/logstash-patterns-core/pull/293/files#diff-5012a418a1738bf6c00a336e452400769dd64001b12ebd22fd14cbc84d7f3b3dR90) has been merged at #293 (part of an effort to ship [ECS compliant captures](https://github.com/logstash-plugins/logstash-patterns-core/issues/278)).

Issue #172, opened 2016-09-15
Pattern TZ does not include India Standard Time (IST)


tithar:
From: https://github.com/logstash-plugins/logstash-patterns-core/blob/master/patterns/grok-patterns

Doesn't include IST. 
https://www.timeanddate.com/time/zones/ist

From:
TZ (?:[APMCE][SD]T|UTC)

To:
TZ (?:[APMCEI][SD]T|UTC)

talevy:
nor does it support British Summer Time (BST).

jordansissel:
It's missing lots of timezones because it is continental-US focused for whatever reason.

So, do we try to add all known abbreviations for timezones? Or do we change TZ to be something like `[A-Z]{3,4}` to match any 3-4 character capitalized text?

robin13:
Here is a pretty comprehensive list of timezones:
https://www.timeanddate.com/time/zones/

+1 to the `[A-Z]{3,4}` pattern.

robin13:
I've made a pull request with this change: https://github.com/logstash-plugins/logstash-patterns-core/pull/235 but I'm not 100% sure it's the right thing to do...
Pro: it would now actually match all timezone abbreviations which would more accurately match what one would assume it should do
Con: it is now one to five upper case letters from A to Z... that includes a lot more than actual timezones...

guyboertje:
This is a list of abbrev, `INFO DEBUG ALERT WARN` should not be matched
```
[...]

PR #230, opened 2018-05-29
adds lowercase to TZ pattern

Rock-HOG:
some log entries may show lowercase timezone.


jsvd:
would you mind adding tests for this change? there are many examples of tests in the `spec/` folder

kares:
Thanks for the PR, seems like this has been stale for a while.
Since the PR was opened a discussion went on at https://github.com/logstash-plugins/logstash-patterns-core/issues/172.

Lower-case names seem quite uncommon and instead the issue focuses on matching more time-zones.
I know Ruby would match such strings but that is more of an implementation detail I believe than an actual intent.
Going to close this one in favour of https://github.com/logstash-plugins/logstash-patterns-core/pull/235, if anyone is seeing lower-case TZ identifiers produced please let us know more details (sample logs) on what kind of apps/services produce them.

PR #235, opened 2018-09-11
Made TZ pattern more inclusive to all timezone patterns of the world

robin13:


guyboertje:
https://github.com/logstash-plugins/logstash-patterns-core/pull/230 ??

and include the regex I built with the suggest MIL pattern instead of the `Z` at the end, perhaps?

robin13:
Regarding #230: I've not seen lowercase timezone in the wild and am doubtful if that is really a pattern which should be matched as standard.

kares:
This is looking legit, I am scraping through PRs and see TZ is still an issue:

I do like what Guy suggested at #172 which this issue is aiming to fix: 
`\b(?:((A[CEFKLMNRWZ]?[AMOW]?|AQT|B[NORT]?R?|C[ACEHIKLOSVX]?[AOU]?|DAV|DDU|E[ACEG]?S?|F[EJKN]|G[AEFIMY]?[LM]?|H[K]?|HOV|I[CORS]?[K]?|JAY|J|K[GOR]?|KRA|KUY|L[HIK]|LIG|LIN|M[AEHMPUVY]?[GRW]?|N[CFPRUZ]?|NOV|OMS|ORA|P[EGHKMOWY]?[ONT]?|QYZ|RE|ROT|S[ABCGR]?[EKM]?|SYO|T[FJKLMORV]|TAH|TRU|U[CUYZ]?|ULA|V[EU]|VLA|VOL|VOS|W[AEFG]?|WAK|WAR|WET|WI|Y[AE][KP])(D?ST|D?T|T)|BRA|MEZ|MSD|MSK|WIB|WITA|UTC|ZULU|Z))\b`

which is a bit of a regexp but still won't much 'arbitrary' text .. regarding MIL zones I am actually not sure (would leave those out of the discussion). @robin13 are you still interested in this (it's fine to not be in which case we should discuss whether it's worth keeping this one open) ?

robin13:
@kares I also like the suggestion made by Guy.

jpmondet:
Coming to get some news about this PR since TZ is still an issue around here (Europe). We recently had a nice outage cause CET and CEST are not matched by the old pattern.

> @kares I also like the suggestion made by Guy.

Would you like us to modify your PR with Guy's suggestion? Or can this PR be merged in this state ? 

Thanks in advance!

PR #237, opened 2018-11-19
Remove duplicate patterns JAVACLASS and JAVAFILE

jsvd:


jsvd:
> maybe some comment in the commit/changelog as to the reasons why those definitions were removed

@robbavey that was my goal with the tests, both test an example of the pattern that only works for the remaining one.

robbavey:
@jsvd Ah cool, ok LGTM then

jakelandis:
@jsvd - thanks for the heads up, issue logged for the ingest node: https://github.com/elastic/elasticsearch/issues/35699

PR #246, opened 2019-08-06
Aws patterns update

jkburges:
Update patterns to more recent versions of log files.

I need a little help to do an end-to-end test with this using the [docker ELK stack](https://github.com/deviantony/docker-elk). Specifically, I'm not sure how to build/install this plugin to that stack.

I'm not sure if we need to maintain backwards compatibility - pretty sure this PR will break for old versions.

There are a bunch of examples for ELB logs in the AWS docs. This PR only tests the "https" one, I have no idea if the other examples will work 🤷‍♂


spec/patterns/aws_spec.rb
```diff
@@ -9,28 +9,29 @@
...
 
-    let(:value) { "2014-02-15T23:39:43.945958Z my-test-loadbalancer 192.168.131.39:2817 10.0.0.1:80 0.000073 0.001048 0.000057 200 200 0 29 \"GET http://www.example.com:80/ HTTP/1.1\"" }
+    let(:value) { "https 2018-07-02T22:23:00.186641Z app/my-loadbalancer/50dc6c495c0c9188 192.168.131.39:2817 10.0.0.1:80 0.086 0.048 0.037 200 200 0 57 \"GET https://www.example.com:443/ HTTP/1.1\" \"curl/7.46.0\" ECDHE-RSA-AES128-GCM-SHA256 TLSv1.2 arn:aws:elasticloadbalancing:us-east-2:123456789012:targetgroup/my-targets/73e2d6bc24d8a067 \"Root=1-58337281-1d84f3d73c47ec4e58577259\" \"www.example.com\" \"arn:aws:acm:us-east-2:123456789012:certificate/12345678-1234-1234-1234-123456789012\" 1 2018-07-02T22:22:48.364000Z \"authenticate,forward\" \"-\" \"-\""}
 
     subject { grok_match(pattern, value) }
 
-    it { should include("timestamp" => "2014-02-15T23:39:43.945958Z" ) }
-    it { should include("elb" => "my-test-loadbalancer" ) }
+    it { should include("type" => "https" ) }
+    it { should include("timestamp" => "2018-07-02T22:23:00.186641Z" ) }
```
jkburges: I updated the test log value completely, based on what is in the AWS docs at the time of writing. Hence some of these existing fields have changed.

kares:
thanks for your efforts, the backwards compatibility breaking is not really an option to merge and ship these.
however, there's an effort on making patterns ECS compliant, and likely most of the concerns raised here should be addressed for the AWS logs at https://github.com/logstash-plugins/logstash-patterns-core/pull/287/files

PR #250, opened 2019-12-02
local part for email address can start with numbers

tehmoon:
Local part for email addresses can start with numbers: https://en.wikipedia.org/wiki/Email_address

It can actually contain any combination but realistically, it be an alphanumeric character.


kares:
Thanks for the PR, going to close this one in favor of a duplicate at https://github.com/logstash-plugins/logstash-patterns-core/pull/250

PR #260, opened 2020-07-01
add HAPROXYERROR pattern to parse error lines.

leandrojmp:
Add the pattern to parse the 'error' lines from HAProxy logs.

From the [documentation](https://cbonte.github.io/haproxy-dconv/2.0/configuration.html#8.2.5), besides the default **http** log line and **tcp** log line, you can also have an **error** log line with the following format.

> Error log format
When an incoming connection fails due to an SSL handshake or an invalid PROXY
protocol header, haproxy will log the event using a shorter, fixed line format.

> Dec  3 18:27:14 localhost  haproxy[6103]: 127.0.0.1:56059 [03/Dec/2012:17:35:10.380] frt/f1:  Connection error during SSL handshake

This error line has the following fields.
```
Field   Format                                Extract from the example above
      1   process_name '[' pid ']:'                             haproxy[6103]:
      2   client_ip ':' client_port                            127.0.0.1:56059
      3   '[' accept_date ']'                       [03/Dec/2012:17:35:10.380]
      4   frontend_name "/" bind_name ":"                              frt/f1:
      5   message                        Connection error during SSL handshake
```

The pattern `%{HAPROXYERROR}` will parse the above line and create two new fields in the output event, `bind_name` and `error_message`, the original `message` field from the description above was changed to `error_message` to avoid conflict with the event `message` field.


leandrojmp:
Is there anything I can do to try to get it merged? 

I can try to write a test based on the [haproxy_spec.rb](https://github.com/logstash-plugins/logstash-patterns-core/blob/master/spec/patterns/haproxy_spec.rb ) file to test the pattern if it is needed, I didn't do that before because I saw that other patterns like `HAPROXYTCP` also do not have a test in the spec.

leandrojmp:
I've just merged my local branch with the current master to resolve the conflict after the directories for `legacy` and `ecs-v1` patterns were created and I've also created the error log pattern using the ecs fields.

PR #262, opened 2020-07-14
Feat: syslog patterns ecs compatibility

kares:
~~Could use a helping :hand: here whether I am following the right path - picked a few questions/remarks bellow ...~~

*HINT: targeting a work-in-progress **ecs-wip** upstream branch - contains a copy of the original patterns for easy diffs.*


patterns/ecs-v1/linux-syslog
```diff
@@ -1,16 +1,16 @@
...
 CRONLOG %{SYSLOGBASE} \(%{USER:user}\) %{CRON_ACTION:action} \(%{DATA:message}\)
 
 SYSLOGLINE %{SYSLOGBASE2} %{GREEDYDATA:message}
 
 # IETF 5424 syslog(8) format (see http://www.rfc-editor.org/info/rfc5424)
-SYSLOG5424PRI <%{NONNEGINT:syslog5424_pri}>
+SYSLOG5424PRI <%{NONNEGINT:log.syslog.facility.code}>
 SYSLOG5424SD \[%{DATA}\]+
-SYSLOG5424BASE %{SYSLOG5424PRI}%{NONNEGINT:syslog5424_ver} +(?:%{TIMESTAMP_ISO8601:syslog5424_ts}|-) +(?:%{IPORHOST:syslog5424_host}|-) +(-|%{SYSLOG5424PRINTASCII:syslog5424_app}) +(-|%{SYSLOG5424PRINTASCII:syslog5424_proc}) +(-|%{SYSLOG5424PRINTASCII:syslog5424_msgid}) +(?:%{SYSLOG5424SD:syslog5424_sd}|-|)
+SYSLOG5424BASE %{SYSLOG5424PRI}%{NONNEGINT:log.syslog.version} +(?:%{TIMESTAMP_ISO8601:timestamp}|-) +(?:%{IPORHOST:event.provider}|-) +(-|%{SYSLOG5424PRINTASCII:log.syslog.app_name}) +(-|%{SYSLOG5424PRINTASCII:log.syslog.procid}) +(-|%{SYSLOG5424PRINTASCII:log.syslog.msgid}) +(?:%{SYSLOG5424SD:log.syslog.structured_data}|-|)
```
kares, line 14: these should capture the same fields as the SYSLOGLINE (unlike previously)
\+ the standardized new pieces: `log.syslog.app_name`, `log.syslog.procid` etc.

patterns/ecs-v1/linux-syslog
```diff
@@ -1,16 +1,16 @@
 SYSLOG5424PRINTASCII [!-~]+
 
-SYSLOGBASE2 (?:%{SYSLOGTIMESTAMP:timestamp}|%{TIMESTAMP_ISO8601:timestamp8601}) (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource}+(?: %{SYSLOGPROG}:|)
-SYSLOGPAMSESSION %{SYSLOGBASE} (?=%{GREEDYDATA:message})%{WORD:pam_module}\(%{DATA:pam_caller}\): session %{WORD:pam_session_state} for user %{USERNAME:username}(?: by %{GREEDYDATA:pam_by})?
+SYSLOGBASE2 (?:%{SYSLOGTIMESTAMP:timestamp}|%{TIMESTAMP_ISO8601:timestamp8601}) (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:event.provider}+(?: %{SYSLOGPROG}:|)
+SYSLOGPAMSESSION %{SYSLOGBASE} (?=%{GREEDYDATA:message})%{WORD:log.syslog.pam.module}\(%{DATA:log.syslog.pam.caller}\): session %{WORD:log.syslog.pam.session_state} for user %{USERNAME:log.syslog.pam.user}(?: by %{GREEDYDATA})?
```
kares, line 4: not sure how much `SYSLOGPAMSESSION` is used - we likely have to keep it around, 
do we use `log.syslog.pam.` namespace or move it out of the syslog-one `log.pam.` (PAM uses syslog)?

given a sample PAM message:
`Jul 14 13:36:03 precision pkexec: pam_unix(polkit-1:session): session opened for user root by (uid=1001)`
the `pam_by` capture seemed 'unfinished' (`"pam_by": "(uid=1001)"`) so I left it out, do we attempt to change parsing or is it okay to parse as before for ecs?
have intentionally restrained myself from changing any patterns but some

and if we parse it's a bit :weight_lifting: coming up with a good name: 
`parse_by`  -> `log.syslog.pam.authenticator` (and maybe `username`  -> `log.syslog.pam.authanticated_user`)

patterns/ecs-v1/grok-patterns
```diff
@@ -80,16 +80,16 @@ DATESTAMP_EVENTLOG %{YEAR}%{MONTHNUM2}%{MONTHDAY}%{HOUR}%{MINUTE}%{SECOND}
...
-SYSLOGFACILITY <%{NONNEGINT:facility}.%{NONNEGINT:priority}>
+SYSLOGFACILITY <%{NONNEGINT:log.syslog.facility.code}.%{NONNEGINT:log.syslog.priority}>
 HTTPDATE %{MONTHDAY}/%{MONTH}/%{YEAR}:%{TIME} %{INT}
 
 # Shortcuts
 QS %{QUOTEDSTRING}
 
 # Log formats
-SYSLOGBASE %{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{SYSLOGPROG}:
+SYSLOGBASE %{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:event.provider} %{SYSLOGPROG}:
```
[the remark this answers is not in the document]
yaauie: IIRC there is a Syslog input that instantiates a Grok filter to do its heavy-lifting, and I believe it does some magic to re-map. IMO, I'm wary of a user-instantiated filter stomping on reserved fields unless explicitly told to do so.

patterns/ecs-v1/linux-syslog
yaauie: Do these need to be logstash-style field references, or does Grok work magic to handle the nesting when given dot-style nesting?

patterns/ecs-v1/linux-syslog
yaauie: I think this is probably a question for @elasitc/ecs

patterns/ecs-v1/linux-syslog
kares: grok doesn't do any magic so they're just flat for now (with the dots).
until all names are reviewed I am going with dot notation - for sanity :wink: 
these will need to be renamed on a final review `log.syslog.version` -> `[log][syslog][version]`

patterns/ecs-v1/linux-syslog
kares: hmm we can not ping a team cross repo even with the typo fixed // cc @webmat @ebeahan

ebeahan: Thanks for the ping 😄 

The `log.*` field set in ECS is intended for details around the logging mechanism or transport itself.  For example, `log.level` can capture a syslog severity level, `log.syslog.severity` to capture the severity code, etc. Some fields being parsed in this case could be mapped to other top-level ECS field sets (e.g. mapping `username` to `user.name` field under the `user` field set). Some examples of auth log parsing from beats [here](https://github.com/elastic/beats/blob/master/filebeat/module/system/auth/ingest/pipeline.yml) show some possibilities.

kares: Thanks Eric, was running beats as a reference but I do not think LS has anything specific for `auth` logs.
Beats is using `system.auth` namespace (and `system.syslog`) but parses different formats - no special handling for PAM session logs. I guess this is potentially half baked on LS's end.

Anyhow the best way forward is (as suggested) to stop 'abusing' the `log.syslog.*` namespace and do `system.syslog.pam.*` here, I think I will drop the by part and just parse this into:
`Jul 14 13:36:03 precision pkexec: pam_unix(polkit-1:session): session opened for user root by (uid=1001)`
```java
"message"=>"pam_unix(polkit-1:session): session opened for user root by (uid=1001)", 
// atm this really ends up as : 
// "message"=>[
//    "Jul 14 13:36:03 precision pkexec: pam_unix(polkit-1:session): session opened for user root by (uid=1001)", 
//    "pam_unix(polkit-1:session): session opened for user root by (uid=1001)" 
// ]
// for the pattern it's best if we do match into "system.syslog.message" 
// and let the input plugin do the rename/rewrite "system.syslog.message" -> "message"
"user.name"=>"root", 
"process.program"=>"pkexec", 
"system"=>{
  "syslog"=>{
    "pam"=>{"module"=>"pam_unix", "session_state"=>"opened", "origin"=>"polkit-1:session"}
    // (previous) "pam_by" => "(uid=0)" dropped
  }
}, 
"timestamp"=>"Jul 14 13:36:03", 
"event.provider"=>"precision" // will be "host.hostname" due beats compatibility
```

ebeahan: Could `process.name` (ECS) be adopted over `process.program` (non-ECS)?

kares: thanks, yeah sorry about that it was meant to be `process.name` - will update the PR

patterns/ecs-v1/linux-syslog
kares: these are going to be `system.syslog` prefixed e.g. `system.syslog.app_name` 
(for beats compatibility - despite beats not matching any of these)

patterns/ecs-v1/grok-patterns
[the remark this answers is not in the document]
kares: continuing on the `timestamp` topic a bit, from a broader perspective.
most top level patterns capture `timestamp` - in case of syslog we could just use `timestamp` (for both timestamp formats) and let the syslog input handle the coercion and set `@timestamp`.

but there's another way we could approach timestamp - to always properly namespace (and let the user or a related plugin take it from there) : 
- e.g. in case of syslog we would always match timestamps into `system.syslog.timestamp`
  (later coerce -> rename to `@timestamp` - this is how beats approaches this)
- a matched redis log would end up with `event[redis][timestamp]`

wdyt? any preference to either name-spaced timestamps by convention or just use `timestamp` for all?

patterns/ecs-v1/linux-syslog
```diff
@@ -1,16 +1,16 @@
...
 
 # IETF 5424 syslog(8) format (see http://www.rfc-editor.org/info/rfc5424)
-SYSLOG5424PRI <%{NONNEGINT:syslog5424_pri}>
+SYSLOG5424PRI <%{NONNEGINT:[log][syslog][facility][code]}>
 SYSLOG5424SD \[%{DATA}\]+
-SYSLOG5424BASE %{SYSLOG5424PRI}%{NONNEGINT:syslog5424_ver} +(?:%{TIMESTAMP_ISO8601:syslog5424_ts}|-) +(?:%{IPORHOST:syslog5424_host}|-) +(-|%{SYSLOG5424PRINTASCII:syslog5424_app}) +(-|%{SYSLOG5424PRINTASCII:syslog5424_proc}) +(-|%{SYSLOG5424PRINTASCII:syslog5424_msgid}) +(?:%{SYSLOG5424SD:syslog5424_sd}|-|)
+SYSLOG5424BASE %{SYSLOG5424PRI}%{NONNEGINT:[system][syslog][version]} +(?:%{TIMESTAMP_ISO8601:timestamp}|-) +(?:%{IPORHOST:[host][hostname]}|-) +(-|%{SYSLOG5424PRINTASCII:[system][syslog][app_name]}) +(-|%{SYSLOG5424PRINTASCII:[system][syslog][procid]}) +(-|%{SYSLOG5424PRINTASCII:[system][syslog][msgid]}) +(?:%{SYSLOG5424SD:[system][syslog][structured_data]}|-|)
 
-SYSLOG5424LINE %{SYSLOG5424BASE} +%{GREEDYDATA:syslog5424_msg}
+SYSLOG5424LINE %{SYSLOG5424BASE} +%{GREEDYDATA:message}
```
kares, line 16: this is a bit of a 'breaking' change but we likely want all syslog logs to match in a similar way, don't we?

as others (`SYSLOGLINE`) do append to `event["message"]` ... 

an alternative (still breaking backward compat for `SYSLOGLINE`) would be for the message part to always go to `system.syslog.message` (or `system.cron.message`) and allow for a clean rename afterwards - `SYSLOGLINE` sample : 

`Mar 16 00:01:25 evita postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]`
```
  "timestamp"=>"Mar 16 00:01:25",
  "logsource"=>"evita",
  "program"=>"postfix/smtpd", 
  "pid"=>"1713", 
  "message"=>[
    "Mar 16 00:01:25 evita postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]", 
    "connect from camomile.cloud9.net[168.100.1.3]"
  ]
```
an alternative ecs proposal would be to not use `message` and always prefix extracted message sub-part : 
```
  "timestamp"=>"Mar 16 00:01:25", 
  "host"=>{"hostname"=>"evita"}
  "process"=>{"name"=>"postfix/smtpd", "pid"=>"1713"},
  "message"=>"Mar 16 00:01:25 evita postfix/smtpd[1713]: connect from camomile.cloud9.net[168.100.1.3]", 
  "system"=>{
    "syslog"=>{
      "message"=>"connect from camomile.cloud9.net[168.100.1.3]"
    }
  }
```

patterns/ecs-v1/grok-patterns
```diff
@@ -80,16 +80,16 @@ DATESTAMP_EVENTLOG %{YEAR}%{MONTHNUM2}%{MONTHDAY}%{HOUR}%{MINUTE}%{SECOND}
...
-SYSLOGFACILITY <%{NONNEGINT:facility}.%{NONNEGINT:priority}>
+SYSLOGFACILITY <%{NONNEGINT:[log][syslog][facility][code]:int}.%{NONNEGINT:[log][syslog][priority]:int}>
 HTTPDATE %{MONTHDAY}/%{MONTH}/%{YEAR}:%{TIME} %{INT}
 
 # Shortcuts
 QS %{QUOTEDSTRING}
 
 # Log formats
-SYSLOGBASE %{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{SYSLOGPROG}:
+SYSLOGBASE %{SYSLOGTIMESTAMP:timestamp} (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:[host][hostname]} %{SYSLOGPROG}:
```
webmat, line 92: Quick note on `host.name` vs `host.hostname`. This may or may not change how you approach the parsing here.

`host.hostname` is meant to be the hostname as is reported by the system or syslog.

`host.name` is meant to default to the hostname, but be configurable. This helps in situations like managing a fleet of Macs mostly named "MacBook-Pro" :-)

The Elastic Security app uses `host.name` is the primary identifier for hosts. If `host.name` is not populated, events won't show up in SIEM, or there will be problems.

So I think mapping this to `host.hostname` is correct in the context of these Syslog groks.

But perhaps we should mention the subtlety above in some Logstash documentation?

patterns/ecs-v1/linux-syslog
```diff
@@ -1,16 +1,16 @@
 SYSLOG5424PRINTASCII [!-~]+
 
-SYSLOGBASE2 (?:%{SYSLOGTIMESTAMP:timestamp}|%{TIMESTAMP_ISO8601:timestamp8601}) (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource}+(?: %{SYSLOGPROG}:|)
-SYSLOGPAMSESSION %{SYSLOGBASE} (?=%{GREEDYDATA:message})%{WORD:pam_module}\(%{DATA:pam_caller}\): session %{WORD:pam_session_state} for user %{USERNAME:username}(?: by %{GREEDYDATA:pam_by})?
+SYSLOGBASE2 (?:%{SYSLOGTIMESTAMP:timestamp}|%{TIMESTAMP_ISO8601:timestamp}) (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:[host][hostname]}+(?: %{SYSLOGPROG}:|)
```
webmat, line 3: If I remember correctly, the convention is to shove the date into `timestamp8601`, parse it with the date filter, which then populates `@timestamp`. Since `timestamp` is not an ECS field, from ECS' point of view, renaming this field is not necessary.

If you're doing this for other reasons such as aligning various grok patterns to use `timestamp` as an interim field instead, that's also fine. I'm just pointing out that this isn't required per ECS.

patterns/ecs-v1/grok-patterns
```diff
@@ -80,16 +80,16 @@ DATESTAMP_EVENTLOG %{YEAR}%{MONTHNUM2}%{MONTHDAY}%{HOUR}%{MINUTE}%{SECOND}
 # Syslog Dates: Month Day HH:MM:SS
 SYSLOGTIMESTAMP %{MONTH} +%{MONTHDAY} %{TIME}
 PROG [\x21-\x5a\x5c\x5e-\x7e]+
-SYSLOGPROG %{PROG:program}(?:\[%{POSINT:pid}\])?
+SYSLOGPROG %{PROG:[process][name]}(?:\[%{POSINT:[process][pid]:int}\])?
 SYSLOGHOST %{IPORHOST}
-SYSLOGFACILITY <%{NONNEGINT:facility}.%{NONNEGINT:priority}>
+SYSLOGFACILITY <%{NONNEGINT:[log][syslog][facility][code]:int}.%{NONNEGINT:[log][syslog][priority]:int}>
```
webmat, line 85: Thanks for taking the opportunity to coerce to int 👍

patterns/ecs-v1/linux-syslog
```diff
@@ -1,16 +1,16 @@
...
+CRONLOG %{SYSLOGBASE} \(%{USER:[user][name]}\) %{CRON_ACTION:[system][cron][action]} \(%{DATA:message}\)
 
 SYSLOGLINE %{SYSLOGBASE2} %{GREEDYDATA:message}
 
 # IETF 5424 syslog(8) format (see http://www.rfc-editor.org/info/rfc5424)
-SYSLOG5424PRI <%{NONNEGINT:syslog5424_pri}>
+SYSLOG5424PRI <%{NONNEGINT:[log][syslog][facility][code]:int}>
 SYSLOG5424SD \[%{DATA}\]+
-SYSLOG5424BASE %{SYSLOG5424PRI}%{NONNEGINT:syslog5424_ver} +(?:%{TIMESTAMP_ISO8601:syslog5424_ts}|-) +(?:%{IPORHOST:syslog5424_host}|-) +(-|%{SYSLOG5424PRINTASCII:syslog5424_app}) +(-|%{SYSLOG5424PRINTASCII:syslog5424_proc}) +(-|%{SYSLOG5424PRINTASCII:syslog5424_msgid}) +(?:%{SYSLOG5424SD:syslog5424_sd}|-|)
+SYSLOG5424BASE %{SYSLOG5424PRI}%{NONNEGINT:[system][syslog][version]} +(?:%{TIMESTAMP_ISO8601:timestamp}|-) +(?:%{IPORHOST:[host][hostname]}|-) +(-|%{SYSLOG5424PRINTASCII:[system][syslog][app_name]}) +(-|%{SYSLOG5424PRINTASCII:[system][syslog][procid]}) +(-|%{SYSLOG5424PRINTASCII:[system][syslog][msgid]}) +(?:%{SYSLOG5424SD:[system][syslog][structured_data]}|-|)
```
webmat, line 14: I think a few ECS fields could be used here as well. Not 100% sure on all of them, perhaps there's subtleties around 5424, but here's what comes to mind:

* `[system][syslog][app_name]` => `[process][name]`
* `[system][syslog][procid]` => `[process][pid]`
* `[system][syslog][msgid]` => `[event][code]`

patterns/ecs-v1/linux-syslog
webmat: Yeah I also support aligning with Beats and nesting under `system.auth.*` and such, for fields that aren't in ECS.

spec/patterns/syslog_spec.rb
```diff
@@ -4,26 +4,14 @@
...
-
-     it "matches ipv6 in the syslog base pattern" do
-        expect(subject).to match("<174>1 2016-11-14T09:49:23+01:00 2000:6a0:b:315:10:23:4:13 named 2255 - -  info: client 10.23.56.9#63295 (i1.tmg.com): query: i1.tmg.com IN A + (10.23.4.13)")
-     end
-   end
-
   it "matches a simple message with pid" do
-    expect(subject).to match("May 11 15:17:02 meow.soy.se CRON[10973]: pam_unix(cron:session): session opened for user root by (uid=0)")
+    match = grok_match subject, "May 11 15:17:02 meow.soy.se CRON[10973]: pam_unix(cron:session): session opened for user root by (uid=0)"
+    expect(match).to include("pid" => "10973", "program" => "CRON")
```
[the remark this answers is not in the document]
kares: yeah, sorry for the confusion. these are just meant as improvements for the legacy specs.
`ecs_compatibility` mode behavior will follow but we need to merge some spec support code.

patterns/ecs-v1/linux-syslog
kares: thanks, it's mostly only to use one `timestamp` - to align with the conventions with other patterns.
this is [handled](https://github.com/logstash-plugins/logstash-input-syslog/blob/v3.4.4/lib/logstash/inputs/syslog.rb#L293-L294) at the input-syslog plugin (before sending it to a date parser) - won't mind having a single `timestamp` : 
`event.set("timestamp", event.get("timestamp8601")) if event.include?("timestamp8601")`

as this is still a draft having these name-spaced `system.syslog.timestamp` is an open option ...
but I think for users the less work there will be to migrate the better so `timestamp` likely wins that battle.

patterns/ecs-v1/grok-patterns
kares: the ecs type is clear so we *should* definitely do that :wink:

patterns/ecs-v1/grok-patterns
kares: I've setup a draft for the syslog input, that might be a good place to put this so I put :scissors: your notes there as a reminder: https://github.com/logstash-plugins/logstash-input-syslog/pull/63#issuecomment-663156390

patterns/ecs-v1/linux-syslog
kares: Thanks Mat, we definitely want to handle all of that (the syslog input already does most of it except for keeping the original message around).

The discussion I was looking for here is whether LS wants to keep append-ing `message` as its matched,
or rather starts using name-spaced matching (striped out part would go to `system.syslog.message`) as a convention to keep the event structure easier to process.

patterns/ecs-v1/linux-syslog
kares: Thanks Mat, seems I got quite confused here (assuming we already had some of the process fields).
Your suggestion is likely what we want - will investigate the source of my confusion and likely update...

patterns/ecs-v1/linux-syslog
kares: seems I got confused here, there's a new discussion on these fields so I will close this one

patterns/ecs-v1/grok-patterns
```diff
@@ -38,14 +38,15 @@ UNIXPATH (/([\w_%!$@:.,+~-]+|\\.)*)+
...
-# doesn't turn into %XX
+URIHOST %{IPORHOST}(?::%{POSINT})?
+# uripath comes loosely from RFC1738, but mostly from what Firefox doesn't turn into %XX
 URIPATH (?:/[A-Za-z0-9$.+!*'(){},~:;=@#%&_\-]*)+
-#URIPARAM \?(?:[A-Za-z0-9]+(?:=(?:[^&]*))?(?:&(?:[A-Za-z0-9]+(?:=(?:[^&]*))?)?)*)?
-URIPARAM \?[A-Za-z0-9$.+!*'|(){},~@#%&/=:;_?\-\[\]<>]*
-URIPATHPARAM %{URIPATH}(?:%{URIPARAM})?
-URI %{URIPROTO}://(?:%{USER}(?::[^@]*)?@)?(?:%{URIHOST})?(?:%{URIPATHPARAM})?
+#URIQUERY (?:[A-Za-z0-9]+(?:=(?:[^&]*))?(?:&(?:[A-Za-z0-9]+(?:=(?:[^&]*))?)?)*)?
+URIQUERY [A-Za-z0-9$.+!*'|(){},~@#%&/=:;_?\-\[\]<>]*
```
kares, line 45: a rename `URIPARAM` -> `URIQUERY` (including the commented part)

kares: naming was being inconsistent with usage down the line. 
going to need the 'new' `URIQUERY` for some of the other patterns

patterns/ecs-v1/grok-patterns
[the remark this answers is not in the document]
kares: started using `timestamp` as a convention for a field we potentially want to end up as `@timestamp`

kares:
@yaauie when you get a chance, please give it a second :eyes: as we're going to need the URI changes down the line ...

patterns/ecs-v1/grok-patterns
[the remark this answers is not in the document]
webmat: I think converging on this convention of always using `timestamp` would then work great with having one way to handle this `timestamp` => `@timestamp` conversion.

The date parser could default to reading from `timestamp` and output to `@timestamp`, and the docs could build on that convention to explain that convention.

I like it

patterns/ecs-v1/grok-patterns
kares: not testing this in isolation -> (functional) tests will exist in several places as this gets used in other patterns

patterns/ecs-v1/linux-syslog
kares: for the record: convention at this point is that we keep using (append-in to) the `message` field

patterns/ecs-v1/grok-patterns
```diff
@@ -67,7 +68,7 @@ TIME (?!<[0-9])%{HOUR}:%{MINUTE}(?::%{SECOND})(?![0-9])
 DATE_US %{MONTHNUM}[/-]%{MONTHDAY}[/-]%{YEAR}
 DATE_EU %{MONTHDAY}[./-]%{MONTHNUM}[./-]%{YEAR}
 ISO8601_TIMEZONE (?:Z|[+-]%{HOUR}(?::?%{MINUTE}))
-ISO8601_SECOND (?:%{SECOND}|60)
+ISO8601_SECOND %{SECOND}
```
yaauie, line 71: This causes a regression WRT leap seconds.

patterns/ecs-v1/linux-syslog
```diff
@@ -1,16 +1,16 @@
 SYSLOG5424PRINTASCII [!-~]+
 
-SYSLOGBASE2 (?:%{SYSLOGTIMESTAMP:timestamp}|%{TIMESTAMP_ISO8601:timestamp8601}) (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource}+(?: %{SYSLOGPROG}:|)
-SYSLOGPAMSESSION %{SYSLOGBASE} (?=%{GREEDYDATA:message})%{WORD:pam_module}\(%{DATA:pam_caller}\): session %{WORD:pam_session_state} for user %{USERNAME:username}(?: by %{GREEDYDATA:pam_by})?
+SYSLOGBASE2 (?:%{SYSLOGTIMESTAMP:timestamp}|%{TIMESTAMP_ISO8601:timestamp})(?: %{SYSLOGFACILITY})?(?: %{SYSLOGHOST:[host][hostname]})?(?: %{SYSLOGPROG}:)?
```
yaauie, line 3: :+1: to capturing to single field `timestamp`.

Propbably need to follow up on the semantics of `[host][hostname]` vs `[host][name]` here too.

yaauie (changes requested):
Overall: wow. super thorough and well thought-out WRT to capturing the `[system]` fields.

This introduces a minor regression WRT to [leap seconds](https://en.wikipedia.org/wiki/Leap_second) (e.g., `2016-06-30T23:59:60Z`, which is valid). It looks like the Date filter can't handle these either when configured with `ISO8601` (and Ruby's `Time` doesn't know what to do with leap seconds anyway) so this isn't a _new_ issue per-se, but I think it is worth keeping the existing behaviour here.

patterns/ecs-v1/grok-patterns
kares: should be fine, was a duplication, [SECOND already handles `60`](https://github.com/logstash-plugins/logstash-patterns-core/blob/03adf12e98e533785347ca2d29cdaae90bb36ed6/patterns/ecs-v1/grok-patterns#L65) + there's also a spec guarding this

patterns/ecs-v1/linux-syslog
kares: filebeat is filling [`host.hostname`](https://github.com/elastic/beats/blob/v7.9.2/filebeat/module/system/syslog/test/suse-syslog.log-expected.json#L8) (as opposed to `host.name`) so it made sense to keep that ... at least as a start

PR #263, opened 2020-07-17
Feat: redis patterns ecs compatibility

kares:
... and shared spec helpers to ease testing in (2) different modes.

The shared helpers, when this is merged, are going to be used for other pattern migration specs ...
Did not split these into 2 PRs on purpose so that we have some spec usage in place.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,3 +9,6 @@ if Dir.exist?(logstash_path) && use_logstash_source
   gem 'logstash-core', :path => "#{logstash_path}/logstash-core"
   gem 'logstash-core-plugin-api', :path => "#{logstash_path}/logstash-core-plugin-api"
 end
+
+# TODO till filter grok with ECS support is released :
+gem 'logstash-filter-grok', github: 'kares/logstash-filter-grok', ref: 'ecs-1-support'
--- a/patterns/ecs-v1/redis
+++ b/patterns/ecs-v1/redis
@@ -1,3 +1,3 @@
 REDISTIMESTAMP %{MONTHDAY} %{MONTH} %{TIME}
-REDISLOG \[%{POSINT:pid}\] %{REDISTIMESTAMP:timestamp} \* 
-REDISMONLOG %{NUMBER:timestamp} \[%{INT:database} %{IP:client}:%{NUMBER:port}\] "%{WORD:command}"\s?%{GREEDYDATA:params}
+REDISLOG \[%{POSINT:[process][pid]:int}\] %{REDISTIMESTAMP:timestamp} \*
+REDISMONLOG %{NUMBER:timestamp} \[%{INT:[redis][database][id]} %{IP:[client][ip]}:%{POSINT:[client][port]:int}\] "%{WORD:[redis][command][name]}"\s?%{GREEDYDATA:[redis][command][args]}
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -24,6 +24,35 @@ end
 require "logstash/filters/grok"
 
 module GrokHelpers
+  module PatternModeSupport
+    @@pattern_mode = nil
+    def pattern_mode
+      @@pattern_mode
+    end
+    module_function :pattern_mode
+
+    def pattern_mode=(mode)
+      @@pattern_mode = mode
+    end
+  end
+
+  def ecs_compatibility?
+    case ecs_compatibility
+    when :disabled then false
+    when nil then nil
+    else true
+    end
+  end
+
+  def ecs_compatibility
+    case mode = PatternModeSupport.pattern_mode
+    when 'legacy' then :disabled
+    when 'ecs-v1' then :v1
+    when nil then nil
+    else fail "pattern_mode: #{mode.inspect}"
+    end
+  end
+
   def grok_match(label, message, exact_match = false)
     grok  = build_grok(label, exact_match)
     event = build_event(message)
@@ -32,11 +61,10 @@ module GrokHelpers
   end
 
   def build_grok(label, exact_match = false)
-    if exact_match
-      grok = LogStash::Filters::Grok.new("match" => ["message", "^%{#{label}}$"])
-    else
-      grok = LogStash::Filters::Grok.new("match" => ["message", "%{#{label}}"])
-    end
+    grok_opts = { "match" => [ "message", exact_match ? "^%{#{label}}$" : "%{#{label}}" ] }
+    ecs_compat = ecs_compatibility # if not set use the plugin default
+    grok_opts["ecs_compatibility"] = ecs_compat unless ecs_compat.nil?
+    grok = LogStash::Filters::Grok.new(grok_opts)
     grok.register
     grok
   end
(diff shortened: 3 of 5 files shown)


Gemfile
```diff
@@ -9,3 +9,6 @@ if Dir.exist?(logstash_path) && use_logstash_source
   gem 'logstash-core', :path => "#{logstash_path}/logstash-core"
   gem 'logstash-core-plugin-api', :path => "#{logstash_path}/logstash-core-plugin-api"
 end
+
+# TODO till filter grok with ECS support is released :
+gem 'logstash-filter-grok', github: 'kares/logstash-filter-grok', ref: 'ecs-1-support'
```
kares, line 14: WiP part but since the gem changes are interdependent we need to start somewhere...

kares:
unfortunately CI is going to be :red_circle: due the `gem ... github:` source declaration :unamused:

spec/spec_helper.rb
```diff
@@ -24,6 +24,35 @@ def pattern_path(path)
 require "logstash/filters/grok"
 
 module GrokHelpers
+  module Shared
```
yaauie, line 27: naming: it is not clear that `Shared` is all about a shared _pattern mode_. Maybe name it `SharedPatternMode` to make it more clear and less likely to become a catch-all for future shared state?

spec/spec_helper.rb
```diff
@@ -48,6 +77,29 @@ def build_event(message)
 
 RSpec.configure do |c|
   c.include GrokHelpers
+  c.include GrokHelpers::Shared
+  c.extend GrokHelpers::Shared
+end
+
+def describe_pattern(name, pattern_modes = [ nil ], &block)
```
yaauie, line 84: I think this method is unnecessarily forgiving about the arguments it accepts, since all current callers send a second array argument we can remove the default value assignment.

```suggestion
/**
 * @param name [String]
 * @param pattern_modes = [Array[String]]
 */
def describe_pattern(name, pattern_modes, &block)
```

spec/spec_helper.rb
```diff
@@ -48,6 +77,29 @@ def build_event(message)
...
+  pattern_modes.each do |mode|
+    RSpec.describe "#{name}#{mode ? " (#{mode})" : nil}" do
+
+      before(:all) do
```
yaauie, line 88: 🤔 I'm not the biggest fan of `before(:all)`, even though it can be a bit more efficient, because it requires ordered non-interleaved execution. Yeah, rspec executes context groups together and doesn't run specs in parallel unless some other testing harness is hooked into, but this "feels" like a sharp edge to me.

spec/spec_helper.rb
```diff
@@ -32,11 +61,11 @@ def grok_match(label, message, exact_match = false)
   end
 
   def build_grok(label, exact_match = false)
-    if exact_match
-      grok = LogStash::Filters::Grok.new("match" => ["message", "^%{#{label}}$"])
-    else
-      grok = LogStash::Filters::Grok.new("match" => ["message", "%{#{label}}"])
+    grok_opts = { "match" => [ "message", exact_match ? "^%{#{label}}$" : "%{#{label}}" ] }
+    if (compat = ecs_compatibility) != nil
```
yaauie, line 65: this setter in the `if` statement makes the code pretty hard to follow. Can we make it more readable?

spec/spec_helper.rb
kares: renamed to `PatternModeSupport`

spec/spec_helper.rb
kares: hopefully cleaner now, the idea was to leave the plugin ecs (legacy) default for existing specs

spec/spec_helper.rb
kares: okay thanks, refactored - no preference really on my end.

spec/spec_helper.rb
kares: this might have been misleading - the idea was to still use `describe_pattern` for stuff that is "left as is",
in which case the plugin just runs in legacy mode (die the `nil` pattern mode). 
updated sample specs here for the `REDISMONLOG` to use `describe_pattern` and thus re-use the grok setup - we could go be explicit about the `legacy` mode there if you have a preference towards explicitness.

have simply anticipated that we might want to keep a lot of specs as is while still re-using pattern/message/grok setup.

patterns/ecs-v1/redis
```diff
@@ -1,3 +1,3 @@
 REDISTIMESTAMP %{MONTHDAY} %{MONTH} %{TIME}
-REDISLOG \[%{POSINT:pid}\] %{REDISTIMESTAMP:timestamp} \* 
-REDISMONLOG %{NUMBER:timestamp} \[%{INT:database} %{IP:client}:%{NUMBER:port}\] "%{WORD:command}"\s?%{GREEDYDATA:params}
+REDISLOG \[%{POSINT:[process][pid]}\] %{REDISTIMESTAMP:timestamp} \*
+REDISMONLOG %{NUMBER:timestamp} \[%{INT:[redis][database][id]} %{IP:[client][ip]}:%{NUMBER:[client][port]}\] "%{WORD:[redis][command][name]}"\s?%{GREEDYDATA:[redis][command][args]}
```
kares, line 3: TODO `client.port` type should be changed to `POSINT` as ECS assumes `type: long` for the field

spec/spec_helper.rb
kares: @yaauie unless you really want us to be explicit I would keep the `pattern_modes` implicit default.
since the plugin is expected to boot in legacy mode, for now ...

patterns/ecs-v1/redis
```diff
@@ -1,3 +1,3 @@
 REDISTIMESTAMP %{MONTHDAY} %{MONTH} %{TIME}
-REDISLOG \[%{POSINT:pid}\] %{REDISTIMESTAMP:timestamp} \* 
-REDISMONLOG %{NUMBER:timestamp} \[%{INT:database} %{IP:client}:%{NUMBER:port}\] "%{WORD:command}"\s?%{GREEDYDATA:params}
+REDISLOG \[%{POSINT:[process][pid]:int}\] %{REDISTIMESTAMP:timestamp} \*
+REDISMONLOG %{NUMBER:timestamp} \[%{INT:[redis][database][id]} %{IP:[client][ip]}:%{POSINT:[client][port]:int}\] "%{WORD:[redis][command][name]}"\s?%{GREEDYDATA:[redis][command][args]}
```
kares, line 3: hey @webmat and @ebeahan we'll be using some custom fields for the redis monitoring command captures:
`"redis" => { "command" => { "name" => "rpush", "args" => "\"my:key\" \"{\\\"data\\\":\"cdr\\\"}\""  } }`

webmat: The ECS fields I see are mapped correctly 👍 

Yes I agree there's no general place to capture queries to storage systems yet, in ECS. This is coming, but there's nothing concrete I can point you to yet. So I agree with using custom fields 👍 

It's worth pointing out that Beats custom fields for the Redis slowlog are slightly different than those you have here. Since these aren't in ECS, I don't mind if you adapt to those or not, but I'm sharing for awareness: https://github.com/elastic/beats/blob/master/filebeat/module/redis/slowlog/_meta/fields.yml

My sense is that we will get general fields to capture queries in ECS before the Stack 8.0 release, so I'll let you decide how you prefer to handle this.

kares: Thanks, did no actually see those as I wasn't really looking at slow-log - assumed this to be a "monitor" log.
There's definitely differences on how LS parses almost everything compared to Beats.
Which is unfortunate and a bit out-of-scope here, anyhow please keep pointing these out anyways :wink: 
This would deserve a `REDISSLOWLOG` pattern but given that Beat's slowlog parsing also feels outdated, I am likely going to keep these as is (reluctant on adding the `redis.slowlog` prefix - since the pattern is named differently).

PR #265, opened 2020-07-20
Feat: ruby and rails pattern ecs compatibility

kares:
The `RUBY_LOGGER` part should be straightforward. 

Rails part could use some specs (after a testing base-line is established: #263)

Here's how `RAILS3` matching looks like before and after:
```
Started GET "/puppet/postfix/notes?target_id=162&target_type=issue&last_fetched_at=1438695732" for 127.0.0.1 at 2015-08-05 07:40:22 +0200
Processing by Projects::NotesController#index as JSON
  Parameters: {"target_id"=>"162", "target_type"=>"issue", "last_fetched_at"=>"1438695732", "namespace_id"=>"puppet", "project_id"=>"postfix"}
  Rendering notes/index.html.erb within layouts/application
  Rendered notes/index.html.erb within layouts/application (1.6ms)
Completed 200 OK in 640ms (Views: 1.7ms | ActiveRecord: 91.0ms)
```

```
  "timestamp"=>"2015-08-05 07:40:22 +0200", 
  "verb"=>"GET", 
  "request"=>"/puppet/postfix/notes?target_id=162&target_type=issue&last_fetched_at=1438695732", 
  "clientip"=>"127.0.0.1", 
  "controller"=>"Projects::NotesController",
  "action"=>"index", 
  "format"=>"JSON", 
  "context"=>"Rendering notes/index.html.erb within layouts/application\n  Rendered notes/index.html.erb within layouts/application (1.6ms)\n", 
  "params"=>"\"target_id\"=>\"162\", \"target_type\"=>\"issue\", \"last_fetched_at\"=>\"1438695732\", \"namespace_id\"=>\"puppet\", \"project_id\"=>\"postfix\"", 
  "response"=>"200"
  "totalms"=>"640", "activerecordms"=>"91.0", "viewms"=>"1.7"
```
```
  "timestamp"=>"2015-08-05 07:40:22 +0200"
  "http"=>{"request"=>{"method"=>"GET"}, "response"=>{"status_code"=>"200"}}
  "url"=>{"original"=>"/puppet/postfix/notes?target_id=162&target_type=issue&last_fetched_at=1438695732"}, 
  "client"=>{"address"=>"127.0.0.1"}, 
  "rails"=>{
    "request"=>{
      "controller"=>{"class"=>"Projects::NotesController", "action"=>"index"},
      "format"=>"JSON",
      "params"=>"\"target_id\"=>\"162\", \"target_type\"=>\"issue\", \"last_fetched_at\"=>\"1438695732\", \"namespace_id\"=>\"puppet\", \"project_id\"=>\"postfix\"", 
      "explain"=>{
        "original"=>"Rendering notes/index.html.erb within layouts/application\n  Rendered notes/index.html.erb within layouts/application (1.6ms)\n"
      }, 
      "duration"=>{"view"=>"1.7", "activerecord"=>"91.0", "total"=>"640"}
    }
  }
```

### DIFF
--- a/patterns/ecs-v1/rails
+++ b/patterns/ecs-v1/rails
@@ -1,13 +1,13 @@
 RUUID \h{32}
 # rails controller with action
-RCONTROLLER (?<controller>[^#]+)#(?<action>\w+)
+RCONTROLLER (?<[rails][controller][class]>[^#]+)#(?<[rails][controller][action]>\w+)
 
 # this will often be the only line:
-RAILS3HEAD (?m)Started %{WORD:verb} "%{URIPATHPARAM:request}" for %{IPORHOST:clientip} at (?<timestamp>%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND} %{ISO8601_TIMEZONE})
+RAILS3HEAD (?m)Started %{WORD:[http][request][method]} "%{URIPATHPARAM:[url][original]}" for %{IPORHOST:[source][address]} at (?<timestamp>%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND} %{ISO8601_TIMEZONE})
 # for some a strange reason, params are stripped of {} - not sure that's a good idea.
-RPROCESSING \W*Processing by %{RCONTROLLER} as (?<format>\S+)(?:\W*Parameters: {%{DATA:params}}\W*)?
-RAILS3FOOT Completed %{NUMBER:response}%{DATA} in %{NUMBER:totalms}ms %{RAILS3PROFILE}%{GREEDYDATA}
-RAILS3PROFILE (?:\(Views: %{NUMBER:viewms}ms \| ActiveRecord: %{NUMBER:activerecordms}ms|\(ActiveRecord: %{NUMBER:activerecordms}ms)?
+RPROCESSING \W*Processing by %{RCONTROLLER} as (?<[rails][request][format]>\S+)(?:\W*Parameters: {%{DATA:[rails][request][params]}}\W*)?
+RAILS3FOOT Completed %{POSINT:[http][response][status_code]:int}%{DATA} in %{NUMBER:[rails][request][duration][total]:float}ms %{RAILS3PROFILE}%{GREEDYDATA}
+RAILS3PROFILE (?:\(Views: %{NUMBER:[rails][request][duration][view]:float}ms \| ActiveRecord: %{NUMBER:[rails][request][duration][active_record]:float}ms|\(ActiveRecord: %{NUMBER:[rails][request][duration][active_record]:float}ms)?
 
 # putting it all together
-RAILS3 %{RAILS3HEAD}(?:%{RPROCESSING})?(?<context>(?:%{DATA}\n)*)(?:%{RAILS3FOOT})?
+RAILS3 %{RAILS3HEAD}(?:%{RPROCESSING})?(?<[rails][request][explain][original]>(?:%{DATA}\n)*)(?:%{RAILS3FOOT})?
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -95,6 +95,7 @@ def describe_pattern(name, pattern_modes = [ nil ], &block)
       let(:pattern) { name }
       let(:message) { raise 'let(:message) { ... } is missing' }
       let(:grok) { grok_match(pattern, message) }
+      subject { grok }
 
       instance_eval(&block)
     end
--- a/patterns/ecs-v1/ruby
+++ b/patterns/ecs-v1/ruby
@@ -1,2 +1,2 @@
 RUBY_LOGLEVEL (?:DEBUG|FATAL|ERROR|WARN|INFO)
-RUBY_LOGGER [DFEWI], \[%{TIMESTAMP_ISO8601:timestamp} #%{POSINT:pid}\] *%{RUBY_LOGLEVEL:loglevel} -- +%{DATA:progname}: %{GREEDYDATA:message}
+RUBY_LOGGER [DFEWI], \[%{TIMESTAMP_ISO8601:timestamp} #%{POSINT:[process][pid]:int}\] *%{RUBY_LOGLEVEL:[log][level]} -- +%{DATA:[process][name]}: %{GREEDYDATA:message}
--- a/spec/patterns/rails3_spec.rb
+++ b/spec/patterns/rails3_spec.rb
@@ -2,17 +2,35 @@
 require "spec_helper"
 require "logstash/patterns/core"
 
-describe_pattern "RAILS3" do
+describe_pattern "RAILS3", ['legacy', 'ecs-v1'] do
 
   context "single-line log" do
 
     let(:message) { 'Started POST "/api/v3/internal/allowed" for 127.0.0.1 at 2015-08-05 11:37:01 +0200' }
 
     # Started
-    it { should include("verb" => "POST" ) }
-    it { should include("request" => "/api/v3/internal/allowed" ) }
+    it do
+      if ecs_compatibility?
+        should include("http" => hash_including("request" => { "method" => "POST" }))
+      else
+        should include("verb" => "POST")
+      end
+    end
+
+    it do
+      if ecs_compatibility?
+      else
+        should include("request" => "/api/v3/internal/allowed")
+      end
+    end
     # for
-    it { should include("clientip" => "127.0.0.1" ) }
+    it do
+      if ecs_compatibility?
+        should include("source" => { "address" => "127.0.0.1" })
+      else
+        should include("clientip" => "127.0.0.1")
+      end
+    end
     # at
     it { should include("timestamp" => "2015-08-05 11:37:01 +0200" ) }
   end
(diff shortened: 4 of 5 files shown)


patterns/ecs-v1/rails
```diff
@@ -1,13 +1,13 @@
...
-RPROCESSING \W*Processing by %{RCONTROLLER} as (?<format>\S+)(?:\W*Parameters: {%{DATA:params}}\W*)?
-RAILS3FOOT Completed %{NUMBER:response}%{DATA} in %{NUMBER:totalms}ms %{RAILS3PROFILE}%{GREEDYDATA}
-RAILS3PROFILE (?:\(Views: %{NUMBER:viewms}ms \| ActiveRecord: %{NUMBER:activerecordms}ms|\(ActiveRecord: %{NUMBER:activerecordms}ms)?
+RPROCESSING \W*Processing by %{RCONTROLLER} as (?<[rails][request][format]>\S+)(?:\W*Parameters: {%{DATA:[rails][request][params]}}\W*)?
+RAILS3FOOT Completed %{NUMBER:[http][response][status_code]}%{DATA} in %{NUMBER:[rails][request][duration][total]}ms %{RAILS3PROFILE}%{GREEDYDATA}
+RAILS3PROFILE (?:\(Views: %{NUMBER:[rails][request][duration][view]}ms \| ActiveRecord: %{NUMBER:[rails][request][duration][activerecord]}ms|\(ActiveRecord: %{NUMBER:[rails][request][duration][activerecord]}ms)?
 
 # putting it all together
-RAILS3 %{RAILS3HEAD}(?:%{RPROCESSING})?(?<context>(?:%{DATA}\n)*)(?:%{RAILS3FOOT})?
+RAILS3 %{RAILS3HEAD}(?:%{RPROCESSING})?(?<[rails][request][explain][original]>(?:%{DATA}\n)*)(?:%{RAILS3FOOT})?
```
kares, line 13: struggled to come up with a better name for the "rendering extra diagnostic data", which is often not present (or ignored).
in theory this could get parsed out by a user further, thus a name-space to put data and `.original` to potentially indicate that they might want to disable indexing of the field.

patterns/ecs-v1/rails
```diff
@@ -1,13 +1,13 @@
 RUUID \h{32}
 # rails controller with action
-RCONTROLLER (?<controller>[^#]+)#(?<action>\w+)
+RCONTROLLER (?<[rails][controller][class]>[^#]+)#(?<[rails][controller][action]>\w+)
 
 # this will often be the only line:
-RAILS3HEAD (?m)Started %{WORD:verb} "%{URIPATHPARAM:request}" for %{IPORHOST:clientip} at (?<timestamp>%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND} %{ISO8601_TIMEZONE})
+RAILS3HEAD (?m)Started %{WORD:[http][request][method]} "%{URIPATHPARAM:[url][original]}" for %{IPORHOST:[client][address]} at (?<timestamp>%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND} %{ISO8601_TIMEZONE})
```
kares, line 6: `client.address` might be the `source.address` in some cases, 
for (often) proxied requests `client.address` made more sense

kares:
cc @webmat @ebeahan if you guys could have a look and see any opportunities here ...
*HINT: it's a draft since it doesn't have tests, yet but otherwise should be good for a review.*

ebeahan:
LGTM 👍

patterns/ecs-v1/rails
webmat: In ECS we want to establish the baseline of always having source/destination populated. Then if the client/server pair makes semantic sense as well, it can be populate too, in addition (not instead of) source/destination.

In the case of web access logs, it's frequent that the destination or server side isn't populated, though. That's fine.

patterns/ecs-v1/rails
```diff
@@ -1,13 +1,13 @@
...
 # for some a strange reason, params are stripped of {} - not sure that's a good idea.
-RPROCESSING \W*Processing by %{RCONTROLLER} as (?<format>\S+)(?:\W*Parameters: {%{DATA:params}}\W*)?
-RAILS3FOOT Completed %{NUMBER:response}%{DATA} in %{NUMBER:totalms}ms %{RAILS3PROFILE}%{GREEDYDATA}
-RAILS3PROFILE (?:\(Views: %{NUMBER:viewms}ms \| ActiveRecord: %{NUMBER:activerecordms}ms|\(ActiveRecord: %{NUMBER:activerecordms}ms)?
+RPROCESSING \W*Processing by %{RCONTROLLER} as (?<[rails][request][format]>\S+)(?:\W*Parameters: {%{DATA:[rails][request][params]}}\W*)?
+RAILS3FOOT Completed %{POSINT:[http][response][status_code]:int}%{DATA} in %{NUMBER:[rails][request][duration][total]:float}ms %{RAILS3PROFILE}%{GREEDYDATA}
```
webmat, line 9: I would map `[rails][request][duration][total]` to `[event][duration]`. Note that you'd have to convert to nanos.

patterns/ecs-v1/rails
webmat: Yes that's a pattern we've used in a few places in ECS as well (under url and user_agent) 👍

patterns/ecs-v1/rails
```diff
@@ -1,13 +1,13 @@
...
+RCONTROLLER (?<[rails][controller][class]>[^#]+)#(?<[rails][controller][action]>\w+)
 
 # this will often be the only line:
-RAILS3HEAD (?m)Started %{WORD:verb} "%{URIPATHPARAM:request}" for %{IPORHOST:clientip} at (?<timestamp>%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND} %{ISO8601_TIMEZONE})
+RAILS3HEAD (?m)Started %{WORD:[http][request][method]} "%{URIPATHPARAM:[url][original]}" for %{IPORHOST:[client][address]} at (?<timestamp>%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND} %{ISO8601_TIMEZONE})
 # for some a strange reason, params are stripped of {} - not sure that's a good idea.
-RPROCESSING \W*Processing by %{RCONTROLLER} as (?<format>\S+)(?:\W*Parameters: {%{DATA:params}}\W*)?
-RAILS3FOOT Completed %{NUMBER:response}%{DATA} in %{NUMBER:totalms}ms %{RAILS3PROFILE}%{GREEDYDATA}
-RAILS3PROFILE (?:\(Views: %{NUMBER:viewms}ms \| ActiveRecord: %{NUMBER:activerecordms}ms|\(ActiveRecord: %{NUMBER:activerecordms}ms)?
+RPROCESSING \W*Processing by %{RCONTROLLER} as (?<[rails][request][format]>\S+)(?:\W*Parameters: {%{DATA:[rails][request][params]}}\W*)?
```
webmat, line 8: Careful with capturing `[rails][request][params]` here. If this is a keyword field I think this is fine. If this is an object field where each param creates new keys, there's a high likelihood of ending up with an Elasticsearch mapping explosion.

patterns/ecs-v1/rails
kares: OK, thanks. That's good to know - going to change this one to `source.address` than ...

patterns/ecs-v1/rails
kares: yeah I did see that use-case in beats (doing the nanos conversion).
however for LS there's no added support "above" the patterns (no `logstash-input-rails_log`), like in beats.
users will have to handle the full ECS "transition" on their own, if they use these grok patterns (except for syslog we have an input there that consumes these).

so I was hesitant to use `event.duration`, think it's better if we do not populate since it expects nanos ...

patterns/ecs-v1/rails
kares: yet again, mappings weren't provided alongside the patterns.
hmm, but now that you mention we might want to provide mapping template snippets... 

from a practical perspective I am a bit hesitant since a lot of the patterns are quite outdated - not sure it's worth the additional effort (as this is a much bigger task than initially assumed already). still, worth a discussion.

patterns/ecs-v1/rails
webmat: Yeah agreed. You shouldn't use event.duration if there's no oppt to convert to nanos

patterns/ecs-v1/rails
webmat: Right. If this is simply a string field that gets the usual dynamic mapping of `text` + `keyword`, no mapping explosion will happen.

spec/spec_helper.rb
```diff
@@ -95,6 +95,7 @@ def describe_pattern(name, pattern_modes = [ nil ], &block)
       let(:pattern) { name }
       let(:message) { raise 'let(:message) { ... } is missing' }
       let(:grok) { grok_match(pattern, message) }
+      subject { grok }
```
yaauie, line 98: nit: named subjects can make the code that uses this easier to read.

In this case, we already are creating a memoized variable, so we could do so with `subject(name)` instead:

``` suggestion
      subject(:grok) { grok_match(pattern, message) }
```

OR, perhaps with a more descriptive name:

``` suggestion
      let(:grok) { grok_match(pattern, message) }
      subject(:grok_result) { grok }
```

patterns/ecs-v1/rails
```diff
@@ -1,13 +1,13 @@
 RUUID \h{32}
 # rails controller with action
-RCONTROLLER (?<controller>[^#]+)#(?<action>\w+)
+RCONTROLLER (?<[rails][controller][class]>[^#]+)#(?<[rails][controller][action]>\w+)
```
yaauie, line 3: I know this is just a carry-forward, but intermixing of literal regex captures and grok-dialect captures makes me twitchy.

spec/spec_helper.rb
kares: thanks put this on the wip branch: https://github.com/logstash-plugins/logstash-patterns-core/commit/f6e13cd8a83c5a62158128808b703bf643516bab

PR #266, opened 2020-07-20
Feat: ecs-ize postgresql named captures

kares:
NOTE: `POSTGRESQL` format seems outdated compared to [beat's](https://github.com/elastic/beats/blob/7.8/filebeat/module/postgresql/log/ingest/pipeline.yml#L7)

### DIFF
--- a/patterns/ecs-v1/postgresql
+++ b/patterns/ecs-v1/postgresql
@@ -1,3 +1,2 @@
 # Default postgresql pg_log format pattern
-POSTGRESQL %{DATESTAMP:timestamp} %{TZ:[event][timezone]} %{DATA:[user][name]} %{GREEDYDATA:[postgresql][log][connection_id]} %{POSINT:[process][pid]}
-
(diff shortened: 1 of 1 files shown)


patterns/ecs-v1/postgresql
```diff
@@ -1,3 +1,3 @@
 # Default postgresql pg_log format pattern
-POSTGRESQL %{DATESTAMP:timestamp} %{TZ} %{DATA:user_id} %{GREEDYDATA:connection_id} %{POSINT:pid}
+POSTGRESQL %{DATESTAMP:timestamp} %{TZ:[event][timezone]} %{DATA:[user][name]} %{GREEDYDATA:[postgresql][log][connection_id]} %{POSINT:[process][pid]}
```
kares, line 2: was going to go with `postgresql.connection.id` 
but due better beats compatibility went with `postgresql.log.connection_id`

NOTE: that beat's `postgresql.log.core_id` is something different ...

kares: any objections going with beats compatibility (despite) the log parsing being different compared to beats?
// cc @webmat @ebeahan

ebeahan: I have no objections.

PR #267, opened 2020-07-21
Feat: ecs-ize httpd patterns

kares:
Beats already has support for [matching](https://github.com/elastic/beats/blob/v7.8.0/filebeat/module/apache/access/ingest/pipeline.yml) */var/log/httpd/access.log* where they're using the `apache.access` prefix for custom fields. 
~~There's no beats support for parsing the Apache error log, have tried using the same prefix convention which turns out a bit confusing e.g. `apache.error.error_code` as noted on the comments inline.~~

Also, yet again, LS' matching patterns seem outdated and could end up matching imperfectly (as seen in the demonstrated new spec). But this is out-of-scope here.

prerequisite: #282

### DIFF
--- a/patterns/ecs-v1/httpd
+++ b/patterns/ecs-v1/httpd
@@ -7,7 +7,7 @@ HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} "(?:-|%{DATA:[http][request][referrer]})" "
 
 # Error logs
 HTTPD20_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{LOGLEVEL:[log][level]}\] (?:\[client %{IPORHOST:[source][address]}\] ){0,1}%{GREEDYDATA:message}
-HTTPD24_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{WORD:[apache][error][module]}:%{LOGLEVEL:[log][level]}\] \[pid %{POSINT:[process][pid]:int}(:tid %{INT:[process][thread][id]:int})?\]( \(%{POSINT:[apache][error][proxy][error_code]?}\)%{DATA:[apache][error][proxy][message]}:)?( \[client %{IPORHOST:[source][address]}:%{POSINT:[source][port]:int}\])?( %{DATA:[apache][error][error_code]}:)? %{GREEDYDATA:message}
+HTTPD24_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{WORD:[apache][error][module]}:%{LOGLEVEL:[log][level]}\] \[pid %{POSINT:[process][pid]:int}(:tid %{INT:[process][thread][id]:int})?\](:? \(%{POSINT:[apache][error][proxy][error_code]?}\)%{DATA:[apache][error][proxy][message]}:)?(:? \[client %{IPORHOST:[source][address]}:%{POSINT:[source][port]:int}\])?(:? %{DATA:[error][code]}:)? %{GREEDYDATA:message}
 HTTPD_ERRORLOG %{HTTPD20_ERRORLOG}|%{HTTPD24_ERRORLOG}
 
 # Deprecated
--- a/patterns/ecs-v1/httpd
+++ b/patterns/ecs-v1/httpd
@@ -2,7 +2,7 @@ HTTPDUSER %{EMAILADDRESS}|%{USER}
 HTTPDERROR_DATE %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
 
 # Log formats
-HTTPD_COMMONLOG %{IPORHOST:[source][address]} %{HTTPDUSER:[apache][access][user][identity]} %{HTTPDUSER:[user][name]} \[%{HTTPDATE:[apache][access][time]}\] "(?:%{WORD:[http][request][method]} %{NOTSPACE:[url][original]}(?: HTTP/%{NUMBER:[http][version]})?|%{DATA})" (?:-|%{NUMBER:[http][response][status_code]}) (?:-|%{NUMBER:[http][response][body][bytes]})
+HTTPD_COMMONLOG %{IPORHOST:[source][address]} %{HTTPDUSER:[apache][access][user][identity]} %{HTTPDUSER:[user][name]} \[%{HTTPDATE:[apache][access][time]}\] "(?:%{WORD:[http][request][method]} %{NOTSPACE:[url][original]}(?: HTTP/%{NUMBER:[http][version]})?|%{DATA})" (?:-|%{INT:[http][response][status_code]}) (?:-|%{INT:[http][response][body][bytes]})
 HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} %{QS:[http][request][referrer]} %{QS:[user_agent][original]}
 
 # Error logs
--- a/patterns/ecs-v1/httpd
+++ b/patterns/ecs-v1/httpd
@@ -2,12 +2,12 @@ HTTPDUSER %{EMAILADDRESS}|%{USER}
 HTTPDERROR_DATE %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
 
 # Log formats
-HTTPD_COMMONLOG %{IPORHOST:[source][address]} %{HTTPDUSER:[apache][access][user][identity]} %{HTTPDUSER:[user][name]} \[%{HTTPDATE:[apache][access][time]}\] "(?:%{WORD:[http][request][method]} %{NOTSPACE:[url][original]}(?: HTTP/%{NUMBER:[http][version]})?|%{DATA})" (?:-|%{INT:[http][response][status_code]}) (?:-|%{INT:[http][response][body][bytes]})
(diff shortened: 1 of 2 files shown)


patterns/ecs-v1/httpd
```diff
@@ -2,12 +2,12 @@ HTTPDUSER %{EMAILADDRESS}|%{USER}
 HTTPDERROR_DATE %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
 
 # Log formats
-HTTPD_COMMONLOG %{IPORHOST:clientip} %{HTTPDUSER:ident} %{HTTPDUSER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
-HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} %{QS:referrer} %{QS:agent}
+HTTPD_COMMONLOG %{IPORHOST:[source][address]} %{HTTPDUSER:[apache][access][user][identity]} %{HTTPDUSER:[user][name]} \[%{HTTPDATE:[apache][access][time]}\] "(?:%{WORD:[http][request][method]} %{NOTSPACE:[url][original]}(?: HTTP/%{NUMBER:[http][version]})?|%{DATA})" %{NUMBER:[http][response][status_code]} (?:%{NUMBER:[http][response][body][bytes]}|-)
```
kares, line 5: - `apache.access.user.identity` this is extra stuff (beats ignores) that is rarely present/relevant
- dropped raw data matching in ecs mode (`rawrequest`) - wonder how useful if at all that is ...
  guess we shouldn't drop any captures that were matched previously?
- otherwise should match ~ same as beats

patterns/ecs-v1/httpd
```diff
@@ -2,12 +2,12 @@ HTTPDUSER %{EMAILADDRESS}|%{USER}
...
+HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} %{QS:[http][request][referrer]} %{QS:[user_agent][original]}
 
 # Error logs
-HTTPD20_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{LOGLEVEL:loglevel}\] (?:\[client %{IPORHOST:clientip}\] ){0,1}%{GREEDYDATA:message}
-HTTPD24_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{WORD:module}:%{LOGLEVEL:loglevel}\] \[pid %{POSINT:pid}(:tid %{NUMBER:tid})?\]( \(%{POSINT:proxy_errorcode}\)%{DATA:proxy_message}:)?( \[client %{IPORHOST:clientip}:%{POSINT:clientport}\])?( %{DATA:errorcode}:)? %{GREEDYDATA:message}
+HTTPD20_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{LOGLEVEL:[log][level]}\] (?:\[client %{IPORHOST:[source][address]}\] ){0,1}%{GREEDYDATA:message}
+HTTPD24_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{WORD:[apache][error][module]}:%{LOGLEVEL:[log][level]}\] \[pid %{POSINT:[process][pid]}(:tid %{NUMBER:[trace][id]})?\]( \(%{POSINT:[apache][error][proxy][error_code]?}\)%{DATA:[apache][error][proxy][message]}:)?( \[client %{IPORHOST:[source][address]}:%{POSINT:[source][port]}\])?( %{DATA:[apache][error][error_code]}:)? %{GREEDYDATA:message}
```
kares, line 10: `apache.error` prefix isn't that obvious - been using that to follow the beats established convention (`apache.access`)
this is open for discussion - using `apache.log.error.` prefix (and `apache.log.access`?) might have a clearer intent.

patterns/ecs-v1/httpd
kares: anything you guys see here to be improved, in terms of ecs mapping/naming? // cc @webmat @ebeahan

patterns/ecs-v1/httpd
ebeahan: One observation: not sure if `trace.id` is the best fit for the `tid` Thread ID field from the error log.

webmat: Indeed, we do have `process.thread.id` for this 👍

patterns/ecs-v1/httpd
```diff
@@ -2,12 +2,12 @@ HTTPDUSER %{EMAILADDRESS}|%{USER}
 HTTPDERROR_DATE %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
 
 # Log formats
-HTTPD_COMMONLOG %{IPORHOST:clientip} %{HTTPDUSER:ident} %{HTTPDUSER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
-HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} %{QS:referrer} %{QS:agent}
+HTTPD_COMMONLOG %{IPORHOST:[source][address]} %{HTTPDUSER:[apache][access][user][identity]} %{HTTPDUSER:[user][name]} \[%{HTTPDATE:[apache][access][time]}\] "(?:%{WORD:[http][request][method]} %{NOTSPACE:[url][original]}(?: HTTP/%{NUMBER:[http][version]})?|%{DATA})" %{INT:[http][response][status_code]:int} (?:%{INT:[http][response][body][bytes]:int}|-)
```
webmat, line 5: * `[apache][access][time]` weren't you moving the "timestamp to parse" to the temporary `timestamp` field?

Rest of this grok looks good 👍

patterns/ecs-v1/httpd
```diff
@@ -2,12 +2,12 @@ HTTPDUSER %{EMAILADDRESS}|%{USER}
...
+HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} %{QS:[http][request][referrer]} %{QS:[user_agent][original]}
 
 # Error logs
-HTTPD20_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{LOGLEVEL:loglevel}\] (?:\[client %{IPORHOST:clientip}\] ){0,1}%{GREEDYDATA:message}
-HTTPD24_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{WORD:module}:%{LOGLEVEL:loglevel}\] \[pid %{POSINT:pid}(:tid %{NUMBER:tid})?\]( \(%{POSINT:proxy_errorcode}\)%{DATA:proxy_message}:)?( \[client %{IPORHOST:clientip}:%{POSINT:clientport}\])?( %{DATA:errorcode}:)? %{GREEDYDATA:message}
+HTTPD20_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{LOGLEVEL:[log][level]}\] (?:\[client %{IPORHOST:[source][address]}\] ){0,1}%{GREEDYDATA:message}
+HTTPD24_ERRORLOG \[%{HTTPDERROR_DATE:timestamp}\] \[%{WORD:[apache][error][module]}:%{LOGLEVEL:[log][level]}\] \[pid %{POSINT:[process][pid]:int}(:tid %{INT:[process][thread][id]:int})?\]( \(%{POSINT:[apache][error][proxy][error_code]?}\)%{DATA:[apache][error][proxy][message]}:)?( \[client %{IPORHOST:[source][address]}:%{POSINT:[source][port]:int}\])?( %{DATA:[apache][error][error_code]}:)? %{GREEDYDATA:message}
```
webmat, line 10: Hmm there can be two different error codes? The proxy one and the other one? Not sure what to suggest here. The one most likely to be present should go to `error.code`.

patterns/ecs-v1/httpd
kares: sorry for the confusion, initially started with `trace.id` 
but the PR has been updated since [(to use `process.thread.id`)](https://github.com/logstash-plugins/logstash-patterns-core/pull/267/files#diff-94b29fa71f3e85689d2377662517ce71R10)

patterns/ecs-v1/httpd
kares: Thanks - the non-proxy one made more sense to become `error.code`
and since the `proxy.message` is really an error message I renamed those to:
`apache.error.proxy.error.code` and `apache.error.proxy.error.message`

spec/patterns/httpd_spec.rb
```diff
@@ -2,30 +2,51 @@
...
+                            'httpversion' => '1.1',
+                            'response' => '200',
+                            'bytes' => '203023',
+                            'referrer' => '"http://semicomplete.com/presentations/logstash-monitorama-2013/"',
+                            'agent' => '"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/32.0.1700.77 Safari/537.36"'
+                        )
+      end
     end
 
     it "does not capture 'null' fields" do
```
yaauie, line 43: 😩 the spec's description says exactly what it should be doing, but in legacy this was yet another victim of the accidental captures that the ECS version now correctly doesn't capture.

:+1: good to have this behave better moving forward.

yaauie (approved):
LGTM 🎉

Issue #123, opened 2016-01-08
Minor change required for TOMCAT_DATESTAMP in patterns/java


rbjorklin:
Using Ubuntu 14.04 not all timestamps in catalina.out matches the provided regexp. Some timestamps are without timezone. A simple and hopefully correct fix for this would be to make the timezone optional. By changing line 17 in patterns/java from:

    TOMCAT_DATESTAMP 20%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:?%{MINUTE}(?::?%{SECOND}) %{ISO8601_TIMEZONE}

to

    TOMCAT_DATESTAMP 20%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:?%{MINUTE}(?::?%{SECOND}) ?%{ISO8601_TIMEZONE}?

solves this problem locally for me. Notice the two added ?, one for the last space and another for ISO8601_TIMEZONE.

kares:
expected to be addressed by https://github.com/logstash-plugins/logstash-patterns-core/pull/268 honestly I wasn't able to find a log example where the time-zone would have been included (used to be always local or whatever `-Duser.timezone` was set).

PR #244, opened 2019-05-07
Support Tomcat 9 log patterns

mitsu-ko:
Hi,

Log format was changed at Tomcat 9. Sample logs are under followings.

26-Apr-2019 14:05:48.484 INFO [main] org.apache.catalina.startup.VersionLoggerListener.log Server version name:   Apache Tomcat/9.0.19
26-Apr-2019 14:05:48.485 INFO [main] org.apache.catalina.startup.VersionLoggerListener.log Server built:          Apr 12 2019 14:22:48 UTC
26-Apr-2019 14:05:48.485 INFO [main] org.apache.catalina.startup.VersionLoggerListener.log Server version number: 9.0.19.0
26-Apr-2019 14:05:48.485 INFO [main] org.apache.catalina.startup.VersionLoggerListener.log OS Name:               Linux

Attached patch can support the new log format.

regards,

PR #268, opened 2020-07-21
Feat: make java patterns match in an ecs compatible way

kares:
should be straightforward except 2 hiccups:
- Java class-name capture naming (no ECS convention really)
- `message` sub-part wasn't being matched (appended) to `message` but rather to `logmessage`
   which I believe is good but we did not yet decide whether that's a convention to follow
   (in which case it should be followed for all ecs-ized patterns -> namespacing sub-message part)
   
---

resolves https://github.com/logstash-plugins/logstash-patterns-core/issues/123 closing https://github.com/logstash-plugins/logstash-patterns-core/pull/244

### DIFF
--- a/patterns/ecs-v1/java
+++ b/patterns/ecs-v1/java
@@ -4,7 +4,7 @@ JAVAFILE (?:[a-zA-Z$_0-9. -]+)
 #Allow special <init>, <clinit> methods
 JAVAMETHOD (?:(<(?:cl)?init>)|[a-zA-Z$_][a-zA-Z$_0-9]*)
 #Line number is optional in special cases 'Native method' or 'Unknown source'
-JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:class}\.%{JAVAMETHOD:method}\(%{JAVAFILE:file}(?::%{NUMBER:line})?\)
+JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:[java][log][origin][class_name]}\.%{JAVAMETHOD:[log][origin][function]}\(%{JAVAFILE:[log][origin][file][name]}(?::%{INT:[log][origin][file][line]:int})?\)
 # Java Logs
 JAVATHREAD (?:[A-Z]{2}-Processor[\d]+)
 JAVALOGMESSAGE (.*)
@@ -12,6 +12,6 @@ JAVALOGMESSAGE (.*)
 CATALINA_DATESTAMP %{MONTH} %{MONTHDAY}, 20%{YEAR} %{HOUR}:?%{MINUTE}(?::?%{SECOND}) (?:AM|PM)
 # yyyy-MM-dd HH:mm:ss,SSS ZZZ eg: 2014-01-09 17:32:25,527 -0800
 TOMCAT_DATESTAMP 20%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:?%{MINUTE}(?::?%{SECOND}) %{ISO8601_TIMEZONE}
-CATALINALOG %{CATALINA_DATESTAMP:timestamp} %{JAVACLASS:class} %{JAVALOGMESSAGE:logmessage}
+CATALINALOG %{CATALINA_DATESTAMP:timestamp} %{JAVACLASS:[java][log][origin][class][name]} %{JAVALOGMESSAGE:message}
 # 2014-01-09 20:03:28,269 -0800 | ERROR | com.example.service.ExampleService - something compeletely unexpected happened...
-TOMCATLOG %{TOMCAT_DATESTAMP:timestamp} \| %{LOGLEVEL:level} \| %{JAVACLASS:class} - %{JAVALOGMESSAGE:logmessage}
+TOMCATLOG %{TOMCAT_DATESTAMP:timestamp} \| %{LOGLEVEL:[log][level]} \| %{JAVACLASS:[java][log][origin][class][name]} - %{JAVALOGMESSAGE:message}
--- a/patterns/ecs-v1/java
+++ b/patterns/ecs-v1/java
@@ -21,17 +21,15 @@ CATALINA8_LOG %{CATALINA8_DATESTAMP:timestamp} %{LOGLEVEL:[log][level]} \[%{DATA
 CATALINA_DATESTAMP (?:%{CATALINA8_DATESTAMP})|(?:%{CATALINA7_DATESTAMP})
 CATALINALOG (?:%{CATALINA8_LOG})|(?:%{CATALINA7_LOG})
 
-# yyyy-MM-dd HH:mm:ss,SSS ZZZ e.g.: 2014-01-09 17:32:25,527 -0800
-# different from CATALINA_DATESTAMP in Tomcat 4.1/5.0 (since 5.5 it is the same)
-TOMCAT50_DATESTAMP %{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND}(?: %{ISO8601_TIMEZONE})?
-TOMCAT50_LOG %{TOMCAT50_DATESTAMP:timestamp}(?: %{LOGLEVEL:[log][level]})? .*?\[%{DATA:[tomcat][context][name]}\](?:%{JAVACLASS:[java][log][origin][class][name]}: %{JAVAMETHOD:[log][origin][function]}\(.*?\))?(?:: %{JAVALOGMESSAGE:message})?
 # in Tomcat 5.5, 6.0, 7.0 it is the same as catalina.out logging format
-TOMCAT6_LOG %{CATALINA7_LOG}
+TOMCAT7_LOG %{CATALINA7_LOG}
 TOMCAT8_LOG %{CATALINA8_LOG}
 
 # NOTE: a weird log we started with - not sure what TC version this should match out of the box (due the | delimiters)
-TOMCATLEGACY_LOG %{TOMCAT50_DATESTAMP:timestamp} \| %{LOGLEVEL:[log][level]} \| %{JAVACLASS:[java][log][origin][class][name]} - %{JAVALOGMESSAGE:message}
+# TODO: we should simply drop this completely!
+TOMCATLEGACY_DATESTAMP %{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND}(?: %{ISO8601_TIMEZONE})?
+TOMCATLEGACY_LOG %{TOMCATLEGACY_DATESTAMP:timestamp} \| %{LOGLEVEL:[log][level]} \| %{JAVACLASS:[java][log][origin][class][name]} - %{JAVALOGMESSAGE:message}
 
 TOMCAT_DATESTAMP (?:%{CATALINA8_DATESTAMP})|(?:%{CATALINA7_DATESTAMP})|(?:%{TOMCAT50_DATESTAMP})
 
-TOMCATLOG (?:%{TOMCAT8_LOG})|(?:%{TOMCAT6_LOG})|(?:%{TOMCAT50_LOG})|(?:%{TOMCATLEGACY_LOG})
+TOMCATLOG (?:%{TOMCAT8_LOG})|(?:%{TOMCAT7_LOG})|(?:%{TOMCATLEGACY_LOG})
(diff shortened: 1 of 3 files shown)


patterns/ecs-v1/java
```diff
@@ -4,14 +4,14 @@ JAVAFILE (?:[A-Za-z0-9_. -]+)
 #Allow special <init>, <clinit> methods
 JAVAMETHOD (?:(<(?:cl)?init>)|[a-zA-Z$_][a-zA-Z$_0-9]*)
 #Line number is optional in special cases 'Native method' or 'Unknown source'
-JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:class}\.%{JAVAMETHOD:method}\(%{JAVAFILE:file}(?::%{NUMBER:line})?\)
+JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:[log][origin][class][name]}\.%{JAVAMETHOD:[log][origin][function]}\(%{JAVAFILE:[log][origin][file][name]}(?::%{INT:[log][origin][file][line]})?\)
```
kares, line 7: these are all ECS existing fields except for `log.origin.class.name`
have looked into what APM does for Java - using it's own (`error`) custom structure (which makes sense since it never really sends a stack-trace frame but pretty much always a full trace in a single document).

alternatively to avoid clashing with the standard we could use `java.log.origin.class_name` or similar ...

patterns/ecs-v1/java
```diff
@@ -4,14 +4,14 @@ JAVAFILE (?:[A-Za-z0-9_. -]+)
...
 # Java Logs
 JAVATHREAD (?:[A-Z]{2}-Processor[\d]+)
 JAVALOGMESSAGE (.*)
 # MMM dd, yyyy HH:mm:ss eg: Jan 9, 2014 7:13:13 AM
 CATALINA_DATESTAMP %{MONTH} %{MONTHDAY}, 20%{YEAR} %{HOUR}:?%{MINUTE}(?::?%{SECOND}) (?:AM|PM)
 # yyyy-MM-dd HH:mm:ss,SSS ZZZ eg: 2014-01-09 17:32:25,527 -0800
 TOMCAT_DATESTAMP 20%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:?%{MINUTE}(?::?%{SECOND}) %{ISO8601_TIMEZONE}
-CATALINALOG %{CATALINA_DATESTAMP:timestamp} %{JAVACLASS:class} %{JAVALOGMESSAGE:logmessage}
+CATALINALOG %{CATALINA_DATESTAMP:timestamp} %{JAVACLASS:[log][origin][class][name]} %{JAVALOGMESSAGE:message}
```
kares, line 15: as noted (in the PR desc) due lack of convention we're mapping this to `message`
... a likely better alternative would be to use smt like `java.log.message`

spec/patterns/java_spec.rb
```diff
@@ -16,3 +16,31 @@
...
+                                "file"=>"StackTraceExample.java",
+                                "line"=>"42"
+                            })
+  end
+
+  context 'generated file' do
+    let(:message) { '  at org.jruby.RubyMethod$INVOKER$i$call.call(RubyMethod$INVOKER$i$call.gen)' }
+    it "matches" do
+      grok = grok_match(pattern, message, true)
+      # TODO does not match
```
kares, line 38: :yawning_face:

patterns/ecs-v1/java
kares: @webmat do you guys have any input seeing this explicitly as `log.origin.class.name`?
this might obviously be open to potential future clashes with ECS but I am not finding a better name :name_badge: 
would be better to have a common ECS field such as `log.origin.owner.name` since it's not always a class?

happy to move this to the ECS repo later, but I would like to start with smt reasonable here. 
there was a previous attempt to [introduce `log.origin.module`](https://github.com/elastic/ecs/issues/785) (understand the use-case presented there fitted `log.logger`) this is a different use-case - the name might fit here despite potential confusion with Java's module names.

patterns/ecs-v1/java
webmat: If this is the main "message" of this event, it should absolutely go to the `message` field at the root of the event 👍  Think of looking at these logs (or any other logs) via the Elastic Logs UI; you wouldn't want the "message" column to be empty, while the message is present elsewhere in the document :-)

spec/patterns/java_spec.rb
webmat: > 🥱

Better: `- [ ]` (with a space between the braces) gives you an actual checkbox in Github ;-)

patterns/ecs-v1/java
webmat: Yeah I still think the PR you reference makes sense to revisit, despite the latest discussions.

We haven't had time to look into it, so I would recommend a custom field like `java.log.origin.class_name` for now. 

In addition to having a potential conflict with a future version of ECS, adding a custom fields within existing ECS fields also causes confusion. Either users wonder why the `log.origin.class.name` field is not present in the ECS docs; or if ECS adds it later as a different name, now you'd have to populate the ECS field  (e.g. `log.origin.module.name`) with the same value.

When the fields is clearly nested in a tech-specific section of your schema, then I think users have an easier time understanding that this is not (yet) part of ECS.

patterns/ecs-v1/java
kares: thank you, my confusion was (started a discussion elsewhere) in what is a 'better' broader (LS) convention: 
whether to first `message => event.original` (and than simply overwrite the `message` field) 
or assume post rewriting `java.log.message => message` 
... first one makes more sense and is more prevalent in patterns already.

I'm adding a note to revisit grok filter's `overwrite => [ message ]` default in ecs compatibility.

spec/patterns/java_spec.rb
kares: thanks, unfortunately I am still reluctant in "fixing" (changing) pattern definitions here, in context of ECS.
initially, thought this to be addressed "later" - maybe I will start cleaning up the base-line in parallel...

kares: moved to https://github.com/logstash-plugins/logstash-patterns-core/pull/272

patterns/ecs-v1/java
webmat: Don't forget to also rename other instances of `[log][origin][class][name]`, like on this line and on line 17 :-)

kares:
OK, this has progressed but unfortunately trying to add a specs for `CATALINALOG` 
turned out weird since the patterns is pretty much broken (even for old Tomcat 6/7): 
```
     "timestamp"=>"Jul 30, 2020 3:00:21 PM",
     "class"=>"org.apache.catalina.core.StandardEngine",
     "logmessage"=>"start\nINFO: Starting Servlet Engine: Apache Tomcat/6.0.20",
```
will need to decide whether to ECS-ize it broken as is or attempt to fix before hand...

patterns/ecs-v1/java
```diff
@@ -4,14 +4,14 @@ JAVAFILE (?:[A-Za-z0-9_. -]+)
...
 #Line number is optional in special cases 'Native method' or 'Unknown source'
-JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:class}\.%{JAVAMETHOD:method}\(%{JAVAFILE:file}(?::%{NUMBER:line})?\)
+JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:[java][log][origin][class_name]}\.%{JAVAMETHOD:[log][origin][function]}\(%{JAVAFILE:[log][origin][file][name]}(?::%{INT:[log][origin][file][line]:int})?\)
 # Java Logs
 JAVATHREAD (?:[A-Z]{2}-Processor[\d]+)
 JAVALOGMESSAGE (.*)
 # MMM dd, yyyy HH:mm:ss eg: Jan 9, 2014 7:13:13 AM
 CATALINA_DATESTAMP %{MONTH} %{MONTHDAY}, 20%{YEAR} %{HOUR}:?%{MINUTE}(?::?%{SECOND}) (?:AM|PM)
 # yyyy-MM-dd HH:mm:ss,SSS ZZZ eg: 2014-01-09 17:32:25,527 -0800
 TOMCAT_DATESTAMP 20%{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:?%{MINUTE}(?::?%{SECOND}) %{ISO8601_TIMEZONE}
```
yaauie, line 14: in 79-1/2 years this is totally gonna break 😂

patterns/ecs-v1/java
```diff
@@ -4,14 +4,14 @@ JAVAFILE (?:[A-Za-z0-9_. -]+)
 #Allow special <init>, <clinit> methods
 JAVAMETHOD (?:(<(?:cl)?init>)|[a-zA-Z$_][a-zA-Z$_0-9]*)
 #Line number is optional in special cases 'Native method' or 'Unknown source'
-JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:class}\.%{JAVAMETHOD:method}\(%{JAVAFILE:file}(?::%{NUMBER:line})?\)
+JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:[java][log][origin][class_name]}\.%{JAVAMETHOD:[log][origin][function]}\(%{JAVAFILE:[log][origin][file][name]}(?::%{INT:[log][origin][file][line]:int})?\)
```
yaauie, line 7: here we have `[java][log][origin][class_name]`, but later we also propose in the `CATALINALOG` and `TOMCATLOG` proposes to capture to `[java][log][origin][class][name]`

```suggestion
JAVASTACKTRACEPART %{SPACE}at %{JAVACLASS:[java][log][origin][class][name]}\.%{JAVAMETHOD:[log][origin][function]}\(%{JAVAFILE:[log][origin][file][name]}(?::%{INT:[log][origin][file][line]:int})?\)
```

patterns/ecs-v1/java
```diff
@@ -4,14 +4,34 @@ JAVAFILE (?:[a-zA-Z$_0-9. -]+)
...
+# yyyy-MM-dd HH:mm:ss,SSS ZZZ e.g.: 2014-01-09 17:32:25,527 -0800
+# different from CATALINA_DATESTAMP in Tomcat 4.1/5.0 (since 5.5 it is the same)
+TOMCAT50_DATESTAMP %{YEAR}-%{MONTHNUM}-%{MONTHDAY} %{HOUR}:%{MINUTE}:%{SECOND}(?: %{ISO8601_TIMEZONE})?
+TOMCAT50_LOG %{TOMCAT50_DATESTAMP:timestamp}(?: %{LOGLEVEL:[log][level]})? .*?\[%{DATA:[tomcat][context][name]}\](?:%{JAVACLASS:[java][log][origin][class][name]}: %{JAVAMETHOD:[log][origin][function]}\(.*?\))?(?:: %{JAVALOGMESSAGE:message})?
+# in Tomcat 5.5, 6.0, 7.0 it is the same as catalina.out logging format
+TOMCAT6_LOG %{CATALINA7_LOG}
+TOMCAT8_LOG %{CATALINA8_LOG}
+
+# NOTE: a weird log we started with - not sure what TC version this should match out of the box (due the | delimiters)
+TOMCATLEGACY_LOG %{TOMCAT50_DATESTAMP:timestamp} \| %{LOGLEVEL:[log][level]} \| %{JAVACLASS:[java][log][origin][class][name]} - %{JAVALOGMESSAGE:message}
```
kares, line 33: the original `TOMCATLOG` we started with ...

kares:
Since previous review this was still missing specs for the `CATALINALOG`/`TOMCATLOG`.
Adding those lead me to actually trying to support the changes in format since Tomcat 4.x (all the way to 9.0).

The *catalina.out* (default) logging format and the context (a.k.a. tomcat-log) logging format have been chaging at times independently of each other. Let me know if this is too much and we should instead revert to base-line (without tests).

`TOMCATLOG`'s problem really is that I am not sure what kind of (default) log format that matches - seems like it only supported (custom) user configured logging.

patterns/ecs-v1/java
kares: believe we should [drop this completely](https://github.com/kares/logstash-patterns-core/commit/6e6da5efa7400aaaffdd2047a56fbad3ff01dc41#diff-d09cb561ff3020a52febb9b8389f9297c0f7b074a2af6b06e6e3b2783c85836bR29) - not seeing any out-of-box Tomcat installation logging like this.
it was clearly a custom format, let's rather 'only' support default context logging in Tomcat 7/8/9 :grey_question:

kares:
instead of trying to support old legacy Tomcat <= 5.0 logging (no-one will really use) I rather decided to not do so and have TOMCATX_LOG aligned with CATALINAX_LOG: https://github.com/logstash-plugins/logstash-patterns-core/pull/268/commits/6e6da5efa7400aaaffdd2047a56fbad3ff01dc41

looking at old Tomcat logging formats really came out of the need of trying to identify smt `TOMCATLOG` would match all the way down to Tomcat 4.x (NONE of the logging defaults match the legacy patters)

yaauie (approved):
:tada: LGTM

PR #269, opened 2020-07-27
Feat: ecs-ize MongoDB pattern captures

kares:
`MONGO3_LOG` is straight forward, the (legacy) `MONGO_SLOWQUERY` could use some :eyes: 

DRAFT since this will need specs for the new ECS names.

### DIFF
--- a/patterns/ecs-v1/mongodb
+++ b/patterns/ecs-v1/mongodb
@@ -1,7 +1,7 @@
-MONGO_LOG %{SYSLOGTIMESTAMP:timestamp} \[%{WORD:component}\] %{GREEDYDATA:message}
+MONGO_LOG %{SYSLOGTIMESTAMP:timestamp} \[%{WORD:[mongodb][component]}\] %{GREEDYDATA:message}
 MONGO_QUERY \{ (?<={ ).*(?= } ntoreturn:) \}
-MONGO_SLOWQUERY %{WORD} %{MONGO_WORDDASH:database}\.%{MONGO_WORDDASH:collection} %{WORD}: %{MONGO_QUERY:query} %{WORD}:%{NONNEGINT:ntoreturn} %{WORD}:%{NONNEGINT:ntoskip} %{WORD}:%{NONNEGINT:nscanned}.*nreturned:%{NONNEGINT:nreturned}..+ (?<duration>[0-9]+)ms
+MONGO_SLOWQUERY %{WORD:[mongodb][profile][op]} %{MONGO_WORDDASH:[mongodb][database]}\.%{MONGO_WORDDASH:[mongodb][collection]} %{WORD}: %{MONGO_QUERY:[mongodb][query][original]} %{WORD}:%{NONNEGINT:[mongodb][profile][ntoreturn]:int} %{WORD}:%{NONNEGINT:[mongodb][profile][ntoskip]:int} %{WORD}:%{NONNEGINT:[mongodb][profile][nscanned]:int}.*nreturned:%{NONNEGINT:[mongodb][profile][nreturned]:int}..+ (?<[mongodb][profile][duration]:int>[0-9]+)ms
 MONGO_WORDDASH \b[\w-]+\b
 MONGO3_SEVERITY \w
 MONGO3_COMPONENT %{WORD}|-
-MONGO3_LOG %{TIMESTAMP_ISO8601:timestamp} %{MONGO3_SEVERITY:severity} %{MONGO3_COMPONENT:component}%{SPACE}(?:\[%{DATA:context}\])? %{GREEDYDATA:message}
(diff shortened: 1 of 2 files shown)


patterns/ecs-v1/mongodb
```diff
@@ -1,7 +1,7 @@
-MONGO_LOG %{SYSLOGTIMESTAMP:timestamp} \[%{WORD:component}\] %{GREEDYDATA:message}
+MONGO_LOG %{SYSLOGTIMESTAMP:timestamp} \[%{WORD:[mongodb][component]}\] %{GREEDYDATA:message}
 MONGO_QUERY \{ (?<={ ).*(?= } ntoreturn:) \}
-MONGO_SLOWQUERY %{WORD} %{MONGO_WORDDASH:database}\.%{MONGO_WORDDASH:collection} %{WORD}: %{MONGO_QUERY:query} %{WORD}:%{NONNEGINT:ntoreturn} %{WORD}:%{NONNEGINT:ntoskip} %{WORD}:%{NONNEGINT:nscanned}.*nreturned:%{NONNEGINT:nreturned}..+ (?<duration>[0-9]+)ms
+MONGO_SLOWQUERY %{WORD:[mongodb][profile][op]} %{MONGO_WORDDASH:[mongodb][database]}\.%{MONGO_WORDDASH:[mongodb][collection]} %{WORD}: %{MONGO_QUERY:[mongodb][query][original]} %{WORD}:%{NONNEGINT:[mongodb][profile][ntoreturn]:int} %{WORD}:%{NONNEGINT:[mongodb][profile][ntoskip]:int} %{WORD}:%{NONNEGINT:[mongodb][profile][nscanned]:int}.*nreturned:%{NONNEGINT:[mongodb][profile][nreturned]:int}..+ (?<[mongodb][profile][duration]:int>[0-9]+)ms
```
kares, line 3: have tried for captures to match what MongoDB's [system.profile](https://docs.mongodb.com/manual/reference/database-profiler/#example-system-profile-document) returns, but there are differences (since their naming is a bit 'non-descriptive' - wasn't sure whether/how close to follow) :
- `%{[mongodb][database]}\.%{[mongodb][collection]}` should also be matched as one (`profile.ns`) field if we want to follow the MongoDB convention
- `[mongodb][query][original]` is meant to be `profile.command` (structure should be parsed further)
- Mongo's *profile* uses: `profile.op` and `profile.millis` (`[mongodb][profile][duration]`)
  NOTE: `mongodb.profile.op` wasn't matched before

kares: would appreciate any feedback you guys might have related to ^^^ // cc @webmat @ebeahan
basically whether it's worth trying to fit the log data "back-to" MongoDB's :
```
{
   "op" : "query",
   "ns" : "test.report",
   "command" : {
      "find" : "report",
      "filter" : { "a" : { "$lte" : 500 } },
      "lsid" : {
         "id" : UUID("5ccd5b81-b023-41f3-8959-bf99ed696ce9")
      },
      "$db" : "test"
   },
   "cursorid" : 33629063128,
   "keysExamined" : 101,
   "docsExamined" : 101,
   "fromMultiPlanner" : true,
   "numYield" : 2,
   "nreturned" : 101,
   ...
```

webmat: At this time ECS doesn't support capturing DB query details, so almost all of these are fine as custom fields.

The one exception I see is the duration. This looks like a great match for `event.duration`.

Now how to design custom fields is purely up to the implementer, ECS doesn't have a strong stance on this.

What we've seen done (and I think this makes sense) is to store the raw capture of third party event more or less as is, under an appropriate custom namespace. This makes users that are used to the third party tool's field names feel at home. So if that JSON block represents the fields in the Mongo logs, I think nesting those with minimal name changes under `mongodb.*` or `mongodb.profile.*` makes sense indeed.

One caveat I would add to the above is that since ECS is all lowercase and underscores, the only normalization I would suggest is underscoring the MongoDB keys instead of keeping camel case: `keysExamined ` => `mongodb.keys_examined`.

kares: Thanks Mat, great feedback. 
The `event.duration` story is the same as elsewhere - we do not do the nanos conversion thus it isn't used.

Given also your feedback, on following third-party layout, there's an interesting decision to make:

- 'break' the convention for `mongodb.profile.duration` since their `profile.millis` is not that great of a name
  \+ also, poorly, but somehow indicates to the user to consider `* nanos` & moving this field -> `event.duration`
- we rather stick to the convention of parsing the query duration part under `profile.millis`

Any personal preference here?

webmat: These are custom fields, so 100% up to you to go either way.

I do feel like improving on the wording and calling it `mongodb.profile.duration` would be appreciated by users.

kares:
have checked recent (3.6.8) MongoDB for potential adjustments needed to `MONGO_SLOWQUERY` but the slow-query log is no longer a separate thing (slow-queries are logged in the main log and also the details are not ordered) - thus no sense to consider adjusting parts for recent versions. note that the generic `MONGO3_LOG` works as expected.

PR #271, opened 2020-07-27
Feat: ecs-ize MCOLLECTIVE named captures

kares:
NOTE: MCollective is deprecated these days and (as the new spec indicates) matching seems to not have been complete.

### DIFF
--- a/spec/patterns/mcollective_spec.rb
+++ b/spec/patterns/mcollective_spec.rb
@@ -1,0 +1,22 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/patterns/core"
+
+describe "MCOLLECTIVE" do
+
+  let(:pattern) { "MCOLLECTIVE" }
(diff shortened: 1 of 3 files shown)


spec/patterns/mcollective_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+describe "MCOLLECTIVE" do
+
+  let(:pattern) { "MCOLLECTIVE" }
+  let(:value) { "I, [2010-12-29T11:15:32.321744 #11479]  INFO -- : mcollectived:33 The Marionette Collective 1.1.0 started logging at info level" }
+
+  subject { grok_match(pattern, value) }
+
+  it { should include("timestamp"=>"2010-12-29T11:15:32.321744") }
+  it { should include("pid"=>"11479") }
```
yaauie, line 13: spec appears to be for pre-ECS. Do we need to add an ECS-mode spec?

kares: yeah, kind of waiting for https://github.com/logstash-plugins/logstash-patterns-core/pull/263 as it brings in the base testing infrastructure.

PR #158, opened 2016-06-23
local part for email address can start with numbers

junho85:
Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/

first character of email local-part can be number

resolves https://github.com/logstash-plugins/logstash-patterns-core/issues/152

PR #273, opened 2020-07-28
Fix: e-mail address's local part to match according to RFC

### LINKED ISSUE #152: EMAILADDRESS pattern is NOT MATCHED 234234234@qq.com
%{EMAILADDRESS:ident} **NOT MATCHED**   234234234@qq.com
- Version: lastest(2.3.2)
- Operating System:Mac
- Sample Data:234234234@qq.com
- Steps to Reproduce: http://grokconstructor.appspot.com/do/match

kares:
Reviewing existing PRs before the ECS split - this one was requested several times.

origin https://github.com/logstash-plugins/logstash-patterns-core/pull/158
resolves #152

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,4 @@
 ## 4.2.0
-  - Fix: e-mail address's local part to match according to RFC [#273](https://github.com/logstash-plugins/logstash-patterns-core/pull/273)
   - Fix: Java stack trace's JAVAFILE to better match generated names
   - Fix: match Information/INFORMATION in LOGLEVEL [#274](https://github.com/logstash-plugins/logstash-patterns-core/pull/274)
   - Fix: NAGIOS TIMEPERIOD unknown (from/to) field matching [#275](https://github.com/logstash-plugins/logstash-patterns-core/pull/275)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## 4.2.0
+  - Fix: e-mail address's local part to match according to RFC [#273](https://github.com/logstash-plugins/logstash-patterns-core/pull/273)
   - Fix: Java stack trace's JAVAFILE to better match generated names
   - Fix: match Information/INFORMATION in LOGLEVEL [#274](https://github.com/logstash-plugins/logstash-patterns-core/pull/274)
   - Fix: NAGIOS TIMEPERIOD unknown (from/to) field matching [#275](https://github.com/logstash-plugins/logstash-patterns-core/pull/275)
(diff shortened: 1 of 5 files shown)


patterns/grok-patterns
```diff
@@ -1,6 +1,6 @@
 USERNAME [a-zA-Z0-9._-]+
 USER %{USERNAME}
-EMAILLOCALPART [a-zA-Z][a-zA-Z0-9_.+-=:]+
+EMAILLOCALPART [a-zA-Z0-9][a-zA-Z0-9_.+-=:]+
```
yaauie, line 3: The non-quoted form of an e-mail address's local part from RFC5322 (`dot-atom`) without backtracking:

```suggestion
EMAILLOCALPART [a-zA-Z0-9!#$%&'*+\-/=?^_`{|}~]+(?:\.[a-zA-Z0-9!#$%&'*+\-/=?^_`{|}~]+)*
```

Valid e-mail addresses it catches that our current pattern doesn't:

 - `a@example.com`
 - `a{b}@example.com`

Cases it does _not_ catch (correctly) that our current pattern does (incorrectly):

 -  `a..b@example.com`
 - `a:b@example.com`

kares: Thanks, I know it's imperfect - specifically tried to avoid going too far to just keep changes 'small' for the moment.
The simple change was present in 3 distinct PRs already - no one seems to have cared taking it further, yet :wink: 
Anyhow point taken. This will need some more work if it's not longer just adding specs over another one's commit.

patterns/grok-patterns
```diff
@@ -1,6 +1,6 @@
 USERNAME [a-zA-Z0-9._-]+
 USER %{USERNAME}
-EMAILLOCALPART [a-zA-Z0-9!#$%&'*+\-/=?^_`{|}~]+(?:\.[a-zA-Z0-9!#$%&'*+\-/=?^_`{|}~]+)*
+EMAILLOCALPART [a-zA-Z0-9!#$%&'*+\-/=?^_`{|}~]{1,64}(?:\.[a-zA-Z0-9!#$%&'*+\-/=?^_`{|}~]{1,62}){0,63}
```
kares, line 3: @yaauie incorporated your suggestion and added these "tweaks" on top ([local-part is limited to 64 characters](https://tools.ietf.org/html/rfc5321#section-4.5.3.1.1)).
we can not validate the dots '.' version to be < 64 chars but could at least help the engine have one less pattern to potentially blow up looping "forever".

kares: realized this might be considered a regression to be in a non-major release, going to revert and only handle the 64 limit in ECS mode.

kares: or to save some cycles simply re-target the PR to only affect the new ECS pattern set (keeping legacy as is).

yaauie (approved):
LGTM 🌮

kares:
re-targeted and updated to ECS-mode only.

PR #277, opened 2020-08-02
Pattern for Citrix ADC

jantytgat:
Pattern for Citrix ADC appliances.

Tested on live LogStash server.
Spec-file developed based on other spec in folder.
Had issues with running the tests on my Windows development machine though...

I'd love to get comments to improve on this pull request.

### DIFF
--- a/patterns/citrixadc
+++ b/patterns/citrixadc
@@ -1,0 +1,1 @@
+CITRIXADC_LOG %{SYSLOG5424PRI:CitrixAdcSyslog5424Pri} %{DATE:CitrixAdcDate}:%{TIME:CitrixAdcTime}  %{NOTSPACE:CitrixAdcHostname} %{NOTSPACE:CitrixAdcPacketEngine} : %{WORD} %{WORD:CitrixAdcModule} %{WORD:CitrixAdcEventType} %{WORD:CitrixAdcEventId} \d :  %{GREEDYDATA:CitrixAdcMessage}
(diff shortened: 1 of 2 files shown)


jantytgat:
Is there any feedback on this pull request? Is there something I can do @kares ?

kares:
Hey Jan, sorry I am not familiar with Citrix ADC however we recently (LS 7.12) did a major change where we now have 2 sets of [patterns](https://github.com/logstash-plugins/logstash-patterns-core/tree/v4.3.0/patterns). The one you're updating is the "legacy" set, we reviewed all existing patterns to be [ECS](https://www.elastic.co/guide/en/ecs/current/index.html) compliant. 

Sample ECS migration: https://github.com/logstash-plugins/logstash-patterns-core/pull/281/files

There's currently no plan on shipping any legacy patterns without an ECS version (ideally all new work should go towards updating the ECS set).

I understand this might be frustrating esp. since reviewing for ECS compatibility might be more work.
If you do not leverage or plan to use ECS atm the PR won't be merged since we lack the capacity to keep converting contributed patterns in context of ECS (this is currently an issue with all open PRs).

jantytgat:
@kares so you mean I should replace my custom field names with the appropriate standard ECS fields as much as possible?
Some of the fields in there, however, are specific to Citrix ADC. So I guess I won't be able to convert them all into standard fields?

kares:
> so you mean I should replace my custom field names with the appropriate standard ECS fields as much as possible?

yes that is how it's been done for existing patterns supported by LS

> Some of the fields in there, however, are specific to Citrix ADC. So I guess I won't be able to convert them all into standard fields?

that's perfectly fine, and fairly [common](https://github.com/logstash-plugins/logstash-patterns-core/blob/v4.3.1/patterns/ecs-v1/squid#L5) in such cases just use a custom name-space such as `[citrix][adc]` for all fields.
there seems to be no support for Citrix ADC with Beats, if there was we usually strive for a similar field layout.

there's also a spec writing [convention](https://github.com/logstash-plugins/logstash-patterns-core/blob/v4.3.1/spec/patterns/squid_spec.rb#L5), we would also appreciate more tests - log samples if you could provide (as likely the burden of maintaining/supporting the new pattern will be left on the LS team)

PR #279, opened 2020-08-04
normalize HTTPD_COMBINEDLOG matching

kares:
There's a few changes here, namely:
- missing `ident`, user `auth`, `response` (code) fields will no longer be captured
  httpd indicates these nulls as `-`, meaning that there won't be a `'response' => "-"` for such line
- `referrer` and user `agent` fields will be de-quoted:
   previosly: `'referrer' => '"http://semicomplete.com/presentations/logstash-monitorama-2013/"'`
   changes to: `'referrer' => 'http://semicomplete.com/presentations/logstash-monitorama-2013/'`

These changes will be required for (later) ECS mode, the changes in null field matches are less concerning.
However matching `http.request.referrer` and `user_agent.original` should be dequoted to align with ECS.

I am well aware that this is breaking (much more than what we've been [collected](https://github.com/logstash-plugins/logstash-patterns-core/pull/272) [so](https://github.com/logstash-plugins/logstash-patterns-core/pull/274) [far](https://github.com/logstash-plugins/logstash-patterns-core/pull/275) for 4.2.0 - still unreleased).


kares:
replaced by https://github.com/logstash-plugins/logstash-patterns-core/pull/280

PR #280, opened 2020-08-04
normalize HTTPD_COMBINEDLOG matching

kares:
NOTE: these changes are the base-line for ECS-ification (https://github.com/logstash-plugins/logstash-patterns-core/pull/267 will rebase off this) and target *ecs-wip* branch.

Summary of changes :
- missing `ident`, user `auth`, `response` (code) fields will no longer be captured
  httpd indicates these nulls as `-`, meaning that there won't be a `'response' => "-"` for such line
- `referrer` and user `agent` fields will be de-quoted:
   previosly: `'referrer' => '"http://semicomplete.com/presentations/logstash-monitorama-2013/"'`
   changes to: `'referrer' => 'http://semicomplete.com/presentations/logstash-monitorama-2013/'`

These changes will be required for (later) ECS mode, the changes in null field matches are less concerning.
However matching `http.request.referrer` and `user_agent.original` needs to be "de-quoted" to align with ECS.

---

*Initially thought I'll just target this for master to be released "before" ECS is shipped.
Unfortunately, there isn't a version limiter in the grok filter (`s.add_runtime_dependency 'logstash-patterns-core'`) - so even if this was major it would eventually slip into the next LS (7.x) release.*

*Reasoning here is that we rather want changes done in relation to ECS to be shipped in one release so users update to changes such as outlined here while still using the legacy set and than handle the new ECS capture names (and type-casts) as the next step. Seems more work and confusion to me to ship these before hand.*

### DIFF
--- a/patterns/legacy/httpd
+++ b/patterns/legacy/httpd
@@ -2,8 +2,8 @@ HTTPDUSER %{EMAILADDRESS}|%{USER}
 HTTPDERROR_DATE %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
 
 # Log formats
-HTTPD_COMMONLOG %{IPORHOST:clientip} %{HTTPDUSER:ident} %{HTTPDUSER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
-HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} %{QS:referrer} %{QS:agent}
+HTTPD_COMMONLOG %{IPORHOST:clientip} (?:-|%{HTTPDUSER:ident}) (?:-|%{HTTPDUSER:auth}) \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" (?:-|%{NUMBER:response}) (?:-|%{NUMBER:bytes})
+HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} "(?:-|%{DATA:referrer})" "(?:-|%{DATA:agent})"
 
 # Error logs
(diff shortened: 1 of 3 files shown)


patterns/legacy/httpd
```diff
@@ -2,8 +2,8 @@ HTTPDUSER %{EMAILADDRESS}|%{USER}
 HTTPDERROR_DATE %{DAY} %{MONTH} %{MONTHDAY} %{TIME} %{YEAR}
 
 # Log formats
-HTTPD_COMMONLOG %{IPORHOST:clientip} %{HTTPDUSER:ident} %{HTTPDUSER:auth} \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" %{NUMBER:response} (?:%{NUMBER:bytes}|-)
-HTTPD_COMBINEDLOG %{HTTPD_COMMONLOG} %{QS:referrer} %{QS:agent}
+HTTPD_COMMONLOG %{IPORHOST:clientip} (?:-|%{HTTPDUSER:ident}) (?:-|%{HTTPDUSER:auth}) \[%{HTTPDATE:timestamp}\] "(?:%{WORD:verb} %{NOTSPACE:request}(?: HTTP/%{NUMBER:httpversion})?|%{DATA:rawrequest})" (?:-|%{NUMBER:response}) (?:-|%{NUMBER:bytes})
```
yaauie, line 5: I'm not sure making the legacy patterns capture IFF value is not `-` is an appropriate change, and feels "breaking" to me (since users may define pipelines that _depend_ on the value being `-` to represent empty).

kares: okay, thought it would ease transition to have a two step migration process (esp. since there are other legacy patterns that are pretty much useless and a breaking change could make them more "useful").
but if we rather not break anything in the legacy parts (even if we do a major release) I am fine with that.

kares: reverted, have only kept missing (`-`) http response matching in legacy mode: https://github.com/logstash-plugins/logstash-patterns-core/pull/280/commits/088705179ad9b2440283d3bd3f17d2aa552cc132
considering this a fix (previously failed to parse) + it's best to keep the legacy vs ecs patterns matching ~ the same.
a spec is included in the changeset already.

kares:
since this is now reduced to only a simple fix to parse access.log with missing (`-`) response code, I'll move this elsewhere

PR #281, opened 2020-08-06
Feat: ecs-ize BIND9 query log captures

kares:
not just a drop in replacement e.g. `querytype` field is parsed further,
detailed changes should be visible from the spec, but also : 

`"query" => ["test.example.com", "test.example.com"], "querytype" => "A +E(0)K"` previously

`"dns" => { "question" => { "name" => "test.example.com", "type" => 'A', "class" => 'IN' },` 
`"bind" => { "log" => { "question" => { "flags" => '+E(0)K' }`ecs-ified

*HINT: quite a hustle to identify a real-world log this was matching, as there were no previous tests, I am planning to follow-up at least with at least minor compatible updates. There's changes in log format in pretty much every minor Bind 9.x version.*

### DIFF
--- a/patterns/ecs-v1/bind
+++ b/patterns/ecs-v1/bind
@@ -1,3 +1,7 @@
 BIND9_TIMESTAMP %{MONTHDAY}[-]%{MONTH}[-]%{YEAR} %{TIME}
 
-BIND9 %{BIND9_TIMESTAMP:timestamp} queries: %{LOGLEVEL:loglevel}: client %{IP:clientip}#%{POSINT:clientport} \(%{GREEDYDATA:query}\): query: %{GREEDYDATA:query} IN %{GREEDYDATA:querytype} \(%{IP:dns}\)
+BIND9_DNSTYPE (?:A|AAAA|CAA|CDNSKEY|CDS|CERT|CNAME|CSYNC|DLV|DNAME|DNSKEY|DS|HINFO|LOC|MX|NAPTR|NS|NSEC|NSEC3|OPENPGPKEY|PTR|RRSIG|RP|SIG|SMIMEA|SOA|SRV|TSIG|TXT|URI)
+
+# dns.question.class is static - only 'IN' is supported by Bind9
+# bind.log.question.name is expected to be a 'duplicate' (same as the dns.question.name capture)
+BIND9 %{BIND9_TIMESTAMP:timestamp} queries: %{LOGLEVEL:[log][level]}: client %{IP:[client][ip]}#%{POSINT:[client][port]:int} \(%{GREEDYDATA:[bind][log][question][name]}\): query: %{GREEDYDATA:[dns][question][name]} (?<[dns][question][class]>IN) %{BIND9_DNSTYPE:[dns][question][type]}(:? %{DATA:[bind][log][question][flags]})? \(%{IP:[server][ip]}\)
--- a/patterns/ecs-v1/bind
+++ b/patterns/ecs-v1/bind
@@ -1,7 +1,13 @@
 BIND9_TIMESTAMP %{MONTHDAY}[-]%{MONTH}[-]%{YEAR} %{TIME}
 
 BIND9_DNSTYPE (?:A|AAAA|CAA|CDNSKEY|CDS|CERT|CNAME|CSYNC|DLV|DNAME|DNSKEY|DS|HINFO|LOC|MX|NAPTR|NS|NSEC|NSEC3|OPENPGPKEY|PTR|RRSIG|RP|SIG|SMIMEA|SOA|SRV|TSIG|TXT|URI)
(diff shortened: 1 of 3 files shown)


kares:
added a backwards compatible change to the legacy pattern as well to handle at least Bind 9.10/9.11 logging format.
(https://github.com/logstash-plugins/logstash-patterns-core/pull/281/commits/e11a54fe5590a6a43c193ad25782186efd8755c0 could have been a separate PR but things are getting a bit verbose with "too many" PRs)

spec/patterns/bind_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  let(:message) do # Bind 9.10
+    '17-Feb-2018 23:06:56.326 queries: info: client 172.26.0.1#12345 (test.example.com): query: test.example.com IN A +E(0)K (172.26.0.3)'
+  end
+
+  it 'matches' do
+    should include("timestamp" => "17-Feb-2018 23:06:56.326")
+    if ecs_compatibility?
+      should include("log" => hash_including("level" => "info"))
+      should include("client" => { "ip" => "172.26.0.1", "port" => 12345 })
```
kares, line 15: guess, yet again, these should be `source.ip` and `source.port` instead of `client.*`

webmat: In ECS we try to have folks always use source/destination as the baseline, and only populate client/server when it makes sense.

However DNS is one of the examples where client/server makes the most sense, because it's UDP.

If looking at a stream of activity with questions and answers, using source/dest you'd get a hard to follow view like:

source.ip | destination.ip | dns.type | question.name
--|--|--|--
10.42.42.42 | 1.1.1.1 | query | google.com
1.1.1.1 | 10.42.42.42 | answer | google.com

Whereas if we view by client/server we have a more consistent view:

client.ip | server.ip | dns.type | question.name
--|--|--|--
10.42.42.42 | 1.1.1.1 | query | google.com
10.42.42.42 | 1.1.1.1 | answer | google.com


So for that reason I would definitely agree with going directly for client/server in the groks, as you're already doing.

Note: I think bind doesn't give us answers, but I would go with client/server nonetheless. It will make it easier to correlate with other sources of DNS events also populating client/server like Packetbeat, Suricata, Zeek.

patterns/ecs-v1/bind
```diff
@@ -1,3 +1,7 @@
 BIND9_TIMESTAMP %{MONTHDAY}[-]%{MONTH}[-]%{YEAR} %{TIME}
 
-BIND9 %{BIND9_TIMESTAMP:timestamp} queries: %{LOGLEVEL:loglevel}: client %{IP:clientip}#%{POSINT:clientport} \(%{GREEDYDATA:query}\): query: %{GREEDYDATA:query} IN %{GREEDYDATA:querytype} \(%{IP:dns}\)
+BIND9_DNSTYPE (?:A|AAAA|CAA|CDNSKEY|CDS|CERT|CNAME|CSYNC|DLV|DNAME|DNSKEY|DS|HINFO|LOC|MX|NAPTR|NS|NSEC|NSEC3|OPENPGPKEY|PTR|RRSIG|RP|SIG|SMIMEA|SOA|SRV|TSIG|TXT|URI)
+
+# dns.question.class is static - only 'IN' is supported by Bind9
+# bind.log.question.name is expected to be a 'duplicate' (same as the dns.question.name capture)
+BIND9 %{BIND9_TIMESTAMP:timestamp} queries: %{LOGLEVEL:[log][level]}: client(:? @0x(?:[0-9A-Fa-f]+))? %{IP:[client][ip]}#%{POSINT:[client][port]:int} \(%{GREEDYDATA:[bind][log][question][name]}\): query: %{GREEDYDATA:[dns][question][name]} (?<[dns][question][class]>IN) %{BIND9_DNSTYPE:[dns][question][type]}(:? %{DATA:[bind][log][question][flags]})? \(%{IP:[server][ip]}\)
```
yaauie, line 7: should we capture the client memory location? Maybe to `[@metadata][bind][client][memory_address]` so a user can chose to populate the field on the main event after it has been parsed?

yaauie:
Note: Build is failing in setup phase, which should be resolved before we merge.

patterns/ecs-v1/bind
kares: honestly, felt only useful for actual BIND9 server debugging - it is very implementation specific information.
this would mean I need to introduce another top-level pattern for `@0x(?:[0-9A-Fa-f]+)` - as did not want to re-use `BASE16NUM` :wink: (since it unnecessarily matches more) or do a mixed-style named capturing using regexp \<syntax\>.

kares:
did another refactoring to splint out the "relevant" log part as `BIND9_QUERYLOGBASE` all functionality stayed the same

PR #283, opened 2020-08-11
(legacy) EXIM base-line for ECS-ification

kares:
another day another grok pattern story: this time the *exim* (mail server) patterns seem incomplete.
there seems to be partial sub-patterns and experiments with excluding specific messages which are hard to follow.

in their current (partial) form it's pretty much impossible to even come up with meaningful tests (for ECS).
the proposal here tries to establish a useful (but still minimal) `EXIM` line - for matching "mail arrival" logs (**<=**).

hoping to build exim **ECS support on top of this**.

NOTES:
- changes to existing patterns are backwards compatible
- all presented specs are passing
- there's a mix of %{:grok} and \<regexp\> style named captures
  due the effort to eliminate introducing more `EXIM_xxx` (legacy) names
- other options include only having `EXIM` definition for tests (in a separate file)

HINT: targeting *ecs-wip* branch, these would only get released once the ECS work is complete

### DIFF
--- a/patterns/legacy/exim
+++ b/patterns/legacy/exim
@@ -11,6 +11,9 @@ EXIM_MSG_SIZE (S=%{NUMBER:exim_msg_size})
 EXIM_HEADER_ID (id=%{NOTSPACE:exim_header_id})
 EXIM_SUBJECT (T=%{QS:exim_subject})
 
-EXIM_MESSAGE_ARRIVAL %{EXIM_DATE:timestamp} (?:%{EXIM_PID} )?%{EXIM_MSGID:exim_msgid} (?<exim_flags><=) (?:%{EXIM_FLAGS:exim_flags} )?(?<exim_status>[a-z:] )?%{EMAILADDRESS:exim_sender_email}(?: (?:%{EXIM_REMOTE_HOST}|%{EXIM_INTERFACE}|%{EXIM_PROTOCOL}|%{EXIM_MSG_SIZE}|%{EXIM_HEADER_ID}|%{EXIM_SUBJECT}|(?:[A-Za-z]{1,4}=%{NOTSPACE})))*(?: for %{EMAILADDRESS:exim_recipient_email})?
+EXIM_UNKNOWN_FIELD (?:[A-Za-z]{1,4}=%{NOTSPACE})
+EXIM_NAMED_FIELDS (?: (?:%{EXIM_REMOTE_HOST}|%{EXIM_INTERFACE}|%{EXIM_PROTOCOL}|%{EXIM_MSG_SIZE}|%{EXIM_HEADER_ID}|%{EXIM_SUBJECT}|%{EXIM_UNKNOWN_FIELD}))*
+
+EXIM_MESSAGE_ARRIVAL %{EXIM_DATE:timestamp} (?:%{EXIM_PID} )?%{EXIM_MSGID:exim_msgid} (?<exim_flags><=) (?:%{EXIM_FLAGS:exim_flags} )?(?<exim_status>[a-z:] )?%{EMAILADDRESS:exim_sender_email}%{EXIM_NAMED_FIELDS}(?: for %{EMAILADDRESS:exim_recipient_email})?
 
 EXIM %{EXIM_MESSAGE_ARRIVAL}
--- a/patterns/legacy/exim
+++ b/patterns/legacy/exim
@@ -1,7 +1,7 @@
 EXIM_MSGID [0-9A-Za-z]{6}-[0-9A-Za-z]{6}-[0-9A-Za-z]{2}
 EXIM_FLAGS (<=|[-=>*]>|[*]{2}|==)
 EXIM_DATE %{YEAR:exim_year}-%{MONTHNUM:exim_month}-%{MONTHDAY:exim_day} %{TIME:exim_time}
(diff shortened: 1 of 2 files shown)


kares:
@yaauie anything against these to have somehow meaningful `EXIM` mail arrival matching to start with?
otherwise ECS-ification makes little sense.

patterns/legacy/exim
```diff
@@ -1,13 +1,16 @@
...
 EXIM_EXCLUDE_TERMS (Message is frozen|(Start|End) queue run| Warning: | retry time not reached | no (IP address|host name) found for (IP address|host) | unexpected disconnection while reading SMTP command | no immediate delivery: |another process is handling this message)
-EXIM_REMOTE_HOST (H=(%{NOTSPACE:remote_hostname} )?(\(%{NOTSPACE:remote_heloname}\) )?\[%{IP:remote_host}\])
+EXIM_REMOTE_HOST (H=(%{NOTSPACE:remote_hostname} )?(\(%{NOTSPACE:remote_heloname}\) )?\[%{IP:remote_host}\])(?::%{POSINT:remote_port})?
 EXIM_INTERFACE (I=\[%{IP:exim_interface}\](:%{NUMBER:exim_interface_port}))
 EXIM_PROTOCOL (P=%{NOTSPACE:protocol})
 EXIM_MSG_SIZE (S=%{NUMBER:exim_msg_size})
 EXIM_HEADER_ID (id=%{NOTSPACE:exim_header_id})
 EXIM_SUBJECT (T=%{QS:exim_subject})
 
+EXIM_MESSAGE_ARRIVAL %{EXIM_DATE:timestamp} (?:%{EXIM_PID} )?%{EXIM_MSGID:exim_msgid} (?<exim_flags><=) (?:%{EXIM_FLAGS:exim_flags} )?(?<exim_status>[a-z:] )?%{EMAILADDRESS:exim_sender_email}(?: (?:%{EXIM_REMOTE_HOST}|%{EXIM_INTERFACE}|%{EXIM_PROTOCOL}|%{EXIM_MSG_SIZE}|%{EXIM_HEADER_ID}|%{EXIM_SUBJECT}|(?:[A-Za-z]{1,4}=%{NOTSPACE})))*(?: for %{EMAILADDRESS:exim_recipient_email})?
```
yaauie, line 14: The `EXIM_MESSAGE_ARRIVAL` pattern ends up being pretty long, and the (impressive) capture-any-named-field-in-any-order bit makes it a _little_ hard to follow. Below, I've broken it up a bit to make it easier to understand and to inject extensibility points.

```suggestion
EXIM_UNKNOWN_FIELD (?:[A-Za-z]{1,4}=%{NOTSPACE})
EXIM_NAMED_FIELDS (?: (?:%{EXIM_REMOTE_HOST}|%{EXIM_INTERFACE}|%{EXIM_PROTOCOL}|%{EXIM_MSG_SIZE}|%{EXIM_HEADER_ID}|%{EXIM_SUBJECT}|#{EXIM_UNKNOWN_FIELD}))*

EXIM_MESSAGE_ARRIVAL %{EXIM_DATE:timestamp} (?:%{EXIM_PID} )?%{EXIM_MSGID:exim_msgid} (?<exim_flags><=) (?:%{EXIM_FLAGS:exim_flags} )?(?<exim_status>[a-z:] )?%{EMAILADDRESS:exim_sender_email}#{EXIM_NAMED_FIELDS}(?: for %{EMAILADDRESS:exim_recipient_email})?
```

yaauie, line 14: Separately, I'm not _quite_ sure I understand this sequence:

~~~
(?<exim_flags><=) (?:%{EXIM_FLAGS:exim_flags} )?
~~~

To me that reads: _required `<=` sequence_ followed by an optional `EXIM_FLAGS`. This is odd because `EXIM_FLAGS` includes the `<=` message arrival flag, so there _should_ be no other flags? The spec does not seem to indicate that there will be one-and-only-one flag per line, but all of its examples exhibit this behaviour. If multiple flags _are_ expected, can we guarantee that the _first_ flag is this message arrival sequence? Can we guarantee that there will be at most two?

spec/patterns/exim_spec.rb
```diff
@@ -0,0 +1,89 @@
+# encoding: utf-8
+require "spec_helper"
+require "logstash/patterns/core"
+
+describe_pattern 'EXIM' do
```
yaauie, line 5: 🎉🎉🎉🎉🎉🎉🎉

yaauie (commented):
I'm sorry that this one fell off my radar.

🎉 * 💯 for the effort to add specs.

I've added a couple comments, and think we should put a little extra effort into creating clean extensibility points. The longer a grok pattern is, the harder it is to extend it without also breaking something else. Of course, the specs should help with that, but the easier we can make it to register new fields without dealing with other complexities, the better IMHO.

patterns/legacy/exim
kares: it's been a while but it seems like I intentionally left it in - if so totally my bad I did not include a spec.
in some cases there might be multiple flags in a single log entry but I am not finding any examples where that would hold (for message arrival) atm. believe we can remove that and start with just matching one flag `<=` here, for now.

kares: removed the optional `(?:%{EXIM_FLAGS:exim_flags} )?`

PR #285, opened 2020-09-28
HOSTNAME allow match on hostnames that contain an underscore

nicpenning:
Any reason why the HOSTNAME expression does not account for underscores ("_") in a hostname?

I found systems that would not match on this GROK expression because we were using IPORHOST and the host name was something like: HNAM12522_OLDL

The workaround was using the NOTSPACE grok expression.


metalalive:
would this commit be merged to master branch ?
I'm about to create a PR for this, but find out the same change in this already-existing PR.
thanks for reply

kares:
The patterns seems to be trying to follow [RFC 952](https://tools.ietf.org/html/rfc952) and the conventions for host-names such as [this one](https://man7.org/linux/man-pages/man7/hostname.7.html#:~:text=Each%20element%20of%20the%20hostname,9%2C%20and%20the%20hyphen%20(%2D).). 
With these `_` is not allowed as a valid character. Anyone dealing with non standard hostnames can easily override the `HOSTNAME` pattern definition.

nicpenning:
Okay fair enough, thanks!

PR #286, opened 2020-10-13
Feat: ecs-ification of BRO_ patterns

kares:
For context somewhere ~3 years ago Bro was rebranded as Zeek.
LS' bro patterns were matching an old version of logs, some patterns got updated along the way several times.

There's an attempt to handle parsing of old Bro logs as well as 'new' Zeek format:

1. if an (old) Bro pattern changed (before the Zeek rebranding) make sure in ECS mode we always match
2. introduce ZEEK_ variants for all BRO_ patterns for improved matching of recent Zeek logs (these usually do not match the old BRO_ variants)

Each pattern has an example spec of the old matching and the new one, linking those here (instead of pasting sample jsons as they would get a bit long with 4 patterns) : 
- [`BRO_HTTP`](https://github.com/logstash-plugins/logstash-patterns-core/pull/286/files#diff-c74a1abc0c16894e67b0d39f2a670fc10412e8829dd134740a5c385192dcb89bR15-R158) http.log - *a bit "verbose" as there was an existing specs*
- [`BRO_DNS`](https://github.com/logstash-plugins/logstash-patterns-core/pull/286/files#diff-c74a1abc0c16894e67b0d39f2a670fc10412e8829dd134740a5c385192dcb89bR268-R304) dns.log
- [`BRO_CONN`](https://github.com/logstash-plugins/logstash-patterns-core/pull/286/files#diff-c74a1abc0c16894e67b0d39f2a670fc10412e8829dd134740a5c385192dcb89bR373-R411) conn.log
- [`BRO_FILES`](https://github.com/logstash-plugins/logstash-patterns-core/pull/286/files#diff-c74a1abc0c16894e67b0d39f2a670fc10412e8829dd134740a5c385192dcb89bR503-R533) files.log

**NOTE:** Beats supports the full Zeek pattern set (we only support 4 here) [since](https://github.com/elastic/beats/pull/14150) 7.6.0, we strive here for maximum compatibility with field naming in Beats.

*TODO: consider moving ZEEK_ patterns into a separate file, for specs it makes sense to have them in one place.*


patterns/ecs-v1/bro
```diff
@@ -1,13 +1,31 @@
...
+
+# dns.log - the 'new' format (added *zeek.dns.rtt*)
+ZEEK_DNS %{NUMBER:timestamp}\t%{NOTSPACE:[zeek][session_id]}\t%{IP:[source][ip]}\t%{INT:[source][port]:int}\t%{IP:[destination][ip]}\t%{INT:[destination][port]:int}\t%{WORD:[network][transport]}\t(?:-|%{INT:[dns][id]:int})\t(?:-|%{NUMBER:[zeek][dns][rtt]:float})\t(?:-|%{BRO_DATA:[dns][question][name]})\t(?:-|%{INT:[zeek][dns][qclass]:int})\t(?:-|%{BRO_DATA:[zeek][dns][qclass_name]})\t(?:-|%{INT:[zeek][dns][qtype]:int})\t(?:-|%{BRO_DATA:[dns][question][type]})\t(?:-|%{INT:[zeek][dns][rcode]:int})\t(?:-|%{BRO_DATA:[dns][response_code]})\t%{BRO_BOOL:[zeek][dns][AA]}\t%{BRO_BOOL:[zeek][dns][TC]}\t%{BRO_BOOL:[zeek][dns][RD]}\t%{BRO_BOOL:[zeek][dns][RA]}\t%{NONNEGINT:[zeek][dns][Z]:int}\t(?:-|%{BRO_DATA:[zeek][dns][answers]})\t(?:-|%{DATA:[zeek][dns][TTLs]})\t(?:-|%{BRO_BOOL:[zeek][dns][rejected]})
```
kares, line 17: the `dns.id` field is being type-casted `%{INT:[dns][id]:int}` (for Beats compatibility).
ECS prescribes the type for [`dns.id`](https://github.com/elastic/ecs/blob/v1.5.0/generated/ecs/ecs_flat.yml#L1503-L1516) as keyword, thus it should stay a string, any preferences?

kares, line 17: we're keeping `source`/`destination` naming convention even for DNS logs (as opposed to `client`/`server`)

patterns/ecs-v1/bro
```diff
@@ -1,13 +1,31 @@
...
+
+# files.log - new format (2 new fields added at the end)
+ZEEK_FILES_TX_HOSTS (?:-|%{IP:[server][ip]})|(?<[zeek][files][tx_hosts]>%{IP:[server][ip]}(?:[\s,]%{IP})+)
+ZEEK_FILES_RX_HOSTS (?:-|%{IP:[client][ip]})|(?<[zeek][files][rx_hosts]>%{IP:[client][ip]}(?:[\s,]%{IP})+)
```
kares, line 30: we're handling cases when multiple IPs are present: the first IP in the set ends up as `client.ip`,
while the whole field is also matched as `rx_hosts` see the [test](https://github.com/logstash-plugins/logstash-patterns-core/pull/286/files#diff-c74a1abc0c16894e67b0d39f2a670fc10412e8829dd134740a5c385192dcb89bR603-R606)

patterns/ecs-v1/bro
```diff
@@ -1,13 +1,31 @@
...
+# http.log - old format (before the Zeek rename) :
+BRO_HTTP %{NUMBER:timestamp}\t%{NOTSPACE:[zeek][session_id]}\t%{IP:[source][ip]}\t%{INT:[source][port]:int}\t%{IP:[destination][ip]}\t%{INT:[destination][port]:int}\t%{INT:[zeek][http][trans_depth]:int}\t(?:-|%{WORD:[http][request][method]})\t(?:-|%{BRO_DATA:[url][domain]})\t(?:-|%{BRO_DATA:[url][original]})\t(?:-|%{BRO_DATA:[http][request][referrer]})\t(?:-|%{BRO_DATA:[user_agent][original]})\t(?:-|%{NUMBER:[http][request][body][bytes]:int})\t(?:-|%{NUMBER:[http][response][body][bytes]:int})\t(?:-|%{POSINT:[http][response][status_code]:int})\t(?:-|%{DATA:[zeek][http][status_msg]})\t(?:-|%{POSINT:[zeek][http][info_code]:int})\t(?:-|%{DATA:[zeek][http][info_msg]})\t(?:-|%{BRO_DATA:[zeek][http][filename]})\t(?:\(empty\)|%{BRO_DATA:[zeek][http][tags]})\t(?:-|%{BRO_DATA:[url][username]})\t(?:-|%{BRO_DATA:[url][password]})\t(?:-|%{BRO_DATA:[zeek][http][proxied]})\t(?:-|%{BRO_DATA:[zeek][http][orig_fuids]})\t(?:-|%{BRO_DATA:[zeek][http][orig_mime_types]})\t(?:-|%{BRO_DATA:[zeek][http][resp_fuids]})\t(?:-|%{BRO_DATA:[zeek][http][resp_mime_types]})
```
kares, line 8: Beats (for all Zeek modules) copies [`event.id`](https://github.com/elastic/beats/blob/v7.9.2/x-pack/filebeat/module/zeek/http/test/http-json.log-expected.json#L22) from [`zeek.session_id`](https://github.com/elastic/beats/blob/v7.9.2/x-pack/filebeat/module/zeek/http/test/http-json.log-expected.json#L67) ... wonder if we should skip `zeek.session_id` and only set `event.id` here (one confusing part would be that in [`BRO_FILES`](https://github.com/logstash-plugins/logstash-patterns-core/pull/286/files#diff-e6850acc553afe24076eb73346c41481b1bda28ce96f608b55fd2405f6d2a381R26) we have the related connection ids and match it as `zeek.files.session_ids`)

kares:
left potential comments to discuss/clarify as comments inline, please take a look as/if needed // cc @ebeahan @webmat

patterns/ecs-v1/bro
ebeahan: As long as the field is type `keyword` in the index template mapping, I don't believe it being cast to an integer will be a problem. It'll still be indexed as a `keyword`.

patterns/ecs-v1/bro
webmat: Yeah I think it makes sense to keep `zeek.session_id` as a custom field, for the relationship with `zeek.files.session_ids`.

Users can always copy it if they want to map it further to ECS, and better align with Beats.

The other way around doesn't work, though, since `event.id` is meant to be one value, so we couldn't capture file's multiple session IDs with that approach.

So I agree with you on this one.

webmat: Distinct comment on this same pattern. ECS 1.7 (to be released with stack 7.11) will include a `mime_type` field for both request and response. So I think you could migrate the mime_type fields here.

https://www.elastic.co/guide/en/ecs/1.7/ecs-http.html

patterns/ecs-v1/bro
```diff
@@ -1,13 +1,31 @@
...
+ZEEK_HTTP %{NUMBER:timestamp}\t%{NOTSPACE:[zeek][session_id]}\t%{IP:[source][ip]}\t%{INT:[source][port]:int}\t%{IP:[destination][ip]}\t%{INT:[destination][port]:int}\t%{INT:[zeek][http][trans_depth]:int}\t(?:-|%{WORD:[http][request][method]})\t(?:-|%{BRO_DATA:[url][domain]})\t(?:-|%{BRO_DATA:[url][original]})\t(?:-|%{BRO_DATA:[http][request][referrer]})\t(?:-|%{NUMBER:[http][version]})\t(?:-|%{BRO_DATA:[user_agent][original]})\t(?:-|%{BRO_DATA:[zeek][http][origin]})\t(?:-|%{NUMBER:[http][request][body][bytes]:int})\t(?:-|%{NUMBER:[http][response][body][bytes]:int})\t(?:-|%{POSINT:[http][response][status_code]:int})\t(?:-|%{DATA:[zeek][http][status_msg]})\t(?:-|%{POSINT:[zeek][http][info_code]:int})\t(?:-|%{DATA:[zeek][http][info_msg]})\t(?:\(empty\)|%{BRO_DATA:[zeek][http][tags]})\t(?:-|%{BRO_DATA:[url][username]})\t(?:-|%{BRO_DATA:[url][password]})\t(?:-|%{BRO_DATA:[zeek][http][proxied]})\t(?:-|%{BRO_DATA:[zeek][http][orig_fuids]})\t(?:-|%{BRO_DATA:[zeek][http][orig_filenames]})\t(?:-|%{BRO_DATA:[zeek][http][orig_mime_types]})\t(?:-|%{BRO_DATA:[zeek][http][resp_fuids]})\t(?:-|%{BRO_DATA:[zeek][http][resp_filenames]})\t(?:-|%{BRO_DATA:[zeek][http][resp_mime_types]})
```
webmat, line 11: I see mime_types here as well 👀

patterns/ecs-v1/bro
```diff
@@ -1,13 +1,31 @@
...
+
+# conn.log - old bro, also supports 'newer' format (optional *zeek.connection.local_resp* flag) compared to non-ecs mode
+BRO_CONN %{NUMBER:timestamp}\t%{NOTSPACE:[zeek][session_id]}\t%{IP:[source][ip]}\t%{INT:[source][port]:int}\t%{IP:[destination][ip]}\t%{INT:[destination][port]:int}\t%{WORD:[network][transport]}\t(?:-|%{BRO_DATA:[network][protocol]})\t(?:-|%{NUMBER:[zeek][connection][duration]:float})\t(?:-|%{INT:[zeek][connection][orig_bytes]:int})\t(?:-|%{INT:[zeek][connection][resp_bytes]:int})\t(?:-|%{BRO_DATA:[zeek][connection][state]})\t(?:-|%{BRO_BOOL:[zeek][connection][local_orig]})\t(?:(?:-|%{BRO_BOOL:[zeek][connection][local_resp]})\t)?(?:-|%{INT:[zeek][connection][missed_bytes]:int})\t(?:-|%{BRO_DATA:[zeek][connection][history]})\t(?:-|%{INT:[source][packets]:int})\t(?:-|%{INT:[source][bytes]:int})\t(?:-|%{INT:[destination][packets]:int})\t(?:-|%{INT:[destination][bytes]:int})\t(?:\(empty\)|%{BRO_DATA:[zeek][connection][tunnel_parents]})
```
webmat, line 20: I think `[zeek][connection][orig_bytes]` and `[zeek][connection][resp_bytes]` could be mapped to `source.bytes` and `destination.bytes` here.

Actually I see the ECS fields used later in the same pattern. Were the fields in the beginning overlooked?

spec/patterns/bro_spec.rb
```diff
@@ -0,0 +1,611 @@
...
+    let(:message) do
+      # NOTE: Zeek sometimes treats set[ip/string] types as ',' separated other times as ' ' separated
+      "1258867934.558264	F2xow8TIkvHG4Zz41	10.1.1.2 10.1.1.1	192.168.1.105,192.168.1.10,192.168.1.11	-	-	0	-	-	-	0.01234	-	F	-	-	-	0	T	-	-	-	-	-	-	-"
+    end
+
+    it 'matches host(s)' do
+      expect(grok).to include("server"=>{"ip" => '10.1.1.2'}) # tx_host
+      expect(grok).to include("client"=>{"ip" => '192.168.1.105'}) # rx_host
+      expect(grok).to include("zeek"=>{"files" => hash_including("tx_hosts" => '10.1.1.2 10.1.1.1')})
+      expect(grok).to include("zeek"=>{"files" => hash_including("rx_hosts" => '192.168.1.105,192.168.1.10,192.168.1.11')})
```
webmat, line 606: Love that it keeps the full list in the custom field 👍

webmat (approved):
I've noted two minor things you may want to address in comments below. Other than these, this looks good to me 👍

patterns/ecs-v1/bro
kares: Thanks, was only looking at the 1.6 tag so missed that. Updated to use `http.request/response.mime_type`.

patterns/ecs-v1/bro
kares: Wasn't overlooked but its definitely a bit confusing, we follow here what Beats is doing - taking the [later fields (`orig_ip_bytes` and `resp_ip_bytes`)](https://github.com/elastic/beats/blob/v7.9.2/x-pack/filebeat/module/zeek/connection/config/connection.yml#L45-L49) and using those as `source/destination.bytes`.

According to the Zeek [docs](https://docs.zeek.org/en/current/scripts/base/protocols/conn/main.zeek.html#type-Conn::Info) the IP bytes are from the IP packet. While `orig_bytes` and `resp_bytes` are taken from the TCP sequence no. Beats drops the `orig_bytes`, `resp_bytes` fields, for LS we capture them (as `[zeek][connection][orig_bytes]` and `[zeek][connection][resp_bytes]`) for the user to decide what to do with them.

patterns/ecs-v1/bro
yaauie: and when a single IP is present, we intentionally _only_ capture `[client][ip]` (_not_ `[zeek][files][rx_hosts]`)?

patterns/ecs-v1/bro
```diff
@@ -1,13 +1,31 @@
...
+ZEEK_HTTP %{NUMBER:timestamp}\t%{NOTSPACE:[zeek][session_id]}\t%{IP:[source][ip]}\t%{INT:[source][port]:int}\t%{IP:[destination][ip]}\t%{INT:[destination][port]:int}\t%{INT:[zeek][http][trans_depth]:int}\t(?:-|%{WORD:[http][request][method]})\t(?:-|%{BRO_DATA:[url][domain]})\t(?:-|%{BRO_DATA:[url][original]})\t(?:-|%{BRO_DATA:[http][request][referrer]})\t(?:-|%{NUMBER:[http][version]})\t(?:-|%{BRO_DATA:[user_agent][original]})\t(?:-|%{BRO_DATA:[zeek][http][origin]})\t(?:-|%{NUMBER:[http][request][body][bytes]:int})\t(?:-|%{NUMBER:[http][response][body][bytes]:int})\t(?:-|%{POSINT:[http][response][status_code]:int})\t(?:-|%{DATA:[zeek][http][status_msg]})\t(?:-|%{POSINT:[zeek][http][info_code]:int})\t(?:-|%{DATA:[zeek][http][info_msg]})\t(?:\(empty\)|%{BRO_DATA:[zeek][http][tags]})\t(?:-|%{BRO_DATA:[url][username]})\t(?:-|%{BRO_DATA:[url][password]})\t(?:-|%{BRO_DATA:[zeek][http][proxied]})\t(?:-|%{BRO_DATA:[zeek][http][orig_fuids]})\t(?:-|%{BRO_DATA:[zeek][http][orig_filenames]})\t(?:-|%{BRO_DATA:[http][request][mime_type]})\t(?:-|%{BRO_DATA:[zeek][http][resp_fuids]})\t(?:-|%{BRO_DATA:[zeek][http][resp_filenames]})\t(?:-|%{BRO_DATA:[http][response][mime_type]})
```
yaauie, line 11: is there value in referencing the `BRO_BOOL` and `BRO_DATA` types from within the `ZEEK_*` patterns, or should we define them separately?

yaauie (commented):
Again, I am impressed with the thought that went into mapping this all. I've left a couple questions, but overall this looks fantastic 🎉

patterns/ecs-v1/bro
kares: correct - that was the intention here: to not have "duplication" for the common case. 
we always want `[client][ip]` filled and if there are multiple IP fields than also capture them into the *_hosts* field.
logs having multiple IPs seems to be very rare - almost not worth the hustle.

patterns/ecs-v1/bro
kares: thanks - that's actually a good catch. 
makes little sense, we should consider spliting this out into separate files - having *ecs-v1/bro* as well as *ecs-v1/zeek* and introduce `ZEEK_BOOL`/`ZEEK_DATA` helper patterns for the later

kares: pattern definitions are now split-ed into dedicated *patterns/ecs-v1/bro* and *patterns/ecs-v1/zeek* file-sets

PR #290, opened 2020-11-09
Feat: ecs-ify EXIM captures

kares:
following up on https://github.com/logstash-plugins/logstash-patterns-core/pull/283 we know have smt to ECS-ify.

We're only handling a single type of exim's log format: logging (mail) message arrival, simplified log example:
```
2010-09-13 05:00:13 [1487] 1Ov4tU-0000Nz-Rm <= mailling.list@domain.com H=mailhost.domain.com [208.42.54.2]:51792 I=[67.215.162.175]:25 P=esmtps S=21778  T="Daily List: Chameleon" for user@example.com
```

matching as:
```ruby
"process"=>{"pid"=>1487},
"exim"=>{
  "log"=>{
    "message"=>{
      "id"=>"1Ov4tU-0000Nz-Rm", 
      "subject"=>"Daily List: Chameleon", 
      "size"=>21778
    }, 
    "header_id"=>"384a86a39e83be0d9b3a94d1feb3119f@domain.com",
    "sender"=>{"email"=>"mailling.list@domain.com"}, 
    "flags"=>"<=", 
    "recipient"=>{"email"=>"user@example.com"}
  }
}, 
"destination"=>{"ip"=>"67.215.162.175", "port"=>25}, 
"source"=>{"ip"=>"208.42.54.2", "address"=>"mailhost.domain.com", "port"=>51792},
"network"=>{"protocol"=>"esmtps"}
```

Struggled a bit with naming, I've outlined the reasoning bellow in the comments ...


patterns/ecs-v1/exim
```diff
@@ -1,13 +1,25 @@
...
+EXIM_PID \[%{POSINT:[process][pid]:int}\]
 EXIM_QT ((\d+y)?(\d+w)?(\d+d)?(\d+h)?(\d+m)?(\d+s)?)
 EXIM_EXCLUDE_TERMS (Message is frozen|(Start|End) queue run| Warning: | retry time not reached | no (IP address|host name) found for (IP address|host) | unexpected disconnection while reading SMTP command | no immediate delivery: |another process is handling this message)
-EXIM_REMOTE_HOST (H=(%{NOTSPACE:remote_hostname} )?(\(%{NOTSPACE:remote_heloname}\) )?\[%{IP:remote_host}\])
-EXIM_INTERFACE (I=\[%{IP:exim_interface}\](:%{NUMBER:exim_interface_port}))
-EXIM_PROTOCOL (P=%{NOTSPACE:protocol})
-EXIM_MSG_SIZE (S=%{NUMBER:exim_msg_size})
-EXIM_HEADER_ID (id=%{NOTSPACE:exim_header_id})
-EXIM_SUBJECT (T=%{QS:exim_subject})
+EXIM_REMOTE_HOST (H=(\[?%{NOTSPACE:[client][address]}\]? )?(\(%{NOTSPACE:[exim][log][remote_address]}\) )?\[%{IP:[client][ip]}\](?::%{POSINT:[client][port]:int})?)
```
kares, line 13: went with `client.address` here using `source.` sounded like there should be a `destination` field as well ... but maybe I am just overthinking - open to suggestions.

`[exim][log][remote_address]` is meant to be the "raw" unverified host from the smtp HELO record, exim than verifies this host and we end up with a `[client][address]` (optional) + the `[client][ip]`

patterns/ecs-v1/exim
```diff
@@ -1,13 +1,25 @@
...
+EXIM_INTERFACE (I=\[%{IP:[exim][log][host][ip]}\](?::%{NUMBER:[exim][log][host][port]:int}))
```
kares, line 14: potentially could have used `host.ip` here but ECS does not have a `host.port` field thus using prefixed instead.

kares:
// cc @ebeahan @webmat if you guys feel like dropping a review here for us

patterns/ecs-v1/exim
ebeahan: @kares You're correct, `source`/`destination` are typically populated together as a pair. But in an instance when only one of the pair is present, populating one without the other is fine. Likewise, the `client`/`server` fields are usually populated together as a pair.

If only the sender details are included, I'd lean towards populating `source` over `client` for these fields.

patterns/ecs-v1/exim
ebeahan: I'm not very familiar with EXIM logs, but would the local interface IP/port perhaps be considered for the `destination` or `server` fields?

patterns/ecs-v1/exim
kares: thanks, have updated to use `source` / `destination`.

patterns/ecs-v1/exim
kares: thanks, makes sense - spent too much time in EXIM logs assuming remote is "always" 127.0.0.1:25 ...
no reason this wouldn't be deployed on a remote host being accessed through an external IP, which the logs would reflect.
updated the pattern and specs.

webmat (commented):
The current mapping LGTM 👍 

For awareness, there's an ECS RFC in progress for email fields ([RFC 0010](https://github.com/elastic/ecs/blob/master/rfcs/text/0010-email.md)), which would be relevant to some of the fields still nested under `exim.log.*`.

We are hoping to get the RFC mature enough for experimental use (merge at "stage 2") in the next few weeks. Perhaps it's worth waiting until then, and migrate the other fields as well?

patterns/ecs-v1/exim
```diff
@@ -1,13 +1,25 @@
...
-EXIM_FLAGS (<=|[-=>*]>|[*]{2}|==)
-EXIM_DATE %{YEAR:exim_year}-%{MONTHNUM:exim_month}-%{MONTHDAY:exim_day} %{TIME:exim_time}
-EXIM_PID \[%{POSINT}\]
+# <=     message arrival
+# =>     normal message delivery
+# ->     additional address in same delivery
+# *>     delivery suppressed by -N
+# **     delivery failed; address bounced
+# ==     delivery deferred; temporary problem
+EXIM_FLAGS (<=|=>|->|*>|**|==|<>|>>)
```
yaauie, line 8: do the literal asterisks (`*`) need escaping here?
does this grouping need a capture?

```suggestion
EXIM_FLAGS (?:<=|=>|->|\*>|\*\*|==|<>|>>)
```

patterns/legacy/exim
```diff
@@ -11,7 +11,7 @@ EXIM_MSG_SIZE (S=%{NUMBER:exim_msg_size})
 EXIM_HEADER_ID (id=%{NOTSPACE:exim_header_id})
 EXIM_SUBJECT (T=%{QS:exim_subject})
 
-EXIM_UNKNOWN_FIELD (?:[A-Za-z]{1,4}=%{NOTSPACE})
+EXIM_UNKNOWN_FIELD (?:[A-Za-z0-9]{1,4}=%{NOTSPACE})
```
yaauie, line 14: good catch. It looks like the 2013 release of [Exim 4.82](https://git.exim.org/exim.git/blob/exim-4_82:/doc/doc-txt/NewStuff#l129) broke our ability to capture unknown fields with the introduction of `M8S` prefix 🤦🏼

yaauie (commented):
Overall, looking on-track. I've left a few issues in-line.

patterns/ecs-v1/exim
```diff
@@ -1,13 +1,25 @@
...
+EXIM_PID \[%{POSINT:[process][pid]:int}\]
 EXIM_QT ((\d+y)?(\d+w)?(\d+d)?(\d+h)?(\d+m)?(\d+s)?)
 EXIM_EXCLUDE_TERMS (Message is frozen|(Start|End) queue run| Warning: | retry time not reached | no (IP address|host name) found for (IP address|host) | unexpected disconnection while reading SMTP command | no immediate delivery: |another process is handling this message)
-EXIM_REMOTE_HOST (H=(%{NOTSPACE:remote_hostname} )?(\(%{NOTSPACE:remote_heloname}\) )?\[%{IP:remote_host}\])
-EXIM_INTERFACE (I=\[%{IP:exim_interface}\](:%{NUMBER:exim_interface_port}))
-EXIM_PROTOCOL (P=%{NOTSPACE:protocol})
-EXIM_MSG_SIZE (S=%{NUMBER:exim_msg_size})
-EXIM_HEADER_ID (id=%{NOTSPACE:exim_header_id})
-EXIM_SUBJECT (T=%{QS:exim_subject})
+EXIM_REMOTE_HOST (H=(\[?%{NOTSPACE:[source][address]}\]? )?(\(%{NOTSPACE:[exim][log][remote_address]}\) )?\[%{IP:[source][ip]}\](?::%{POSINT:[source][port]:int})?)
```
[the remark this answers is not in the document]
kares: thanks - you're right no examples and I am failing to google any log lines that would have this. thus removing `[]`

patterns/ecs-v1/exim
```diff
@@ -1,13 +1,25 @@
...
+EXIM_INTERFACE (I=\[%{IP:[destination][ip]}\](?::%{NUMBER:[destination][port]:int}))
+EXIM_PROTOCOL (P=%{NOTSPACE:[network][protocol]})
+EXIM_MSG_SIZE (S=%{NUMBER:[exim][log][message][size]:int})
+EXIM_HEADER_ID (id=%{NOTSPACE:[exim][log][header_id]})
+EXIM_SUBJECT (T="%{GREEDYDATA:[exim][log][message][subject]}")
```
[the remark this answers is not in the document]
kares: greedyness was justified here due potential `"` in the capture as demonstrated by the spec:
`T=\"[Examples Galore] Please moderate: \\\"Hello world!\\\"\" for admin@example.net`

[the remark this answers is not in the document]
kares: also the examples listed above are for `=>` "normal message delivery" we only support `<=` "message arrival".
per EXIM documentation: `T           on <= lines: message subject (topic)`, meaning seems unspecified for others.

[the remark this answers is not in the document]
kares: avoiding `GREEDYDATA` leads us to using a look-behind - not sure how much more effective this gets: 
```diff
-EXIM_SUBJECT (T="%{GREEDYDATA:[exim][log][message][subject]}")
+EXIM_QUOTED_CONTENT (?:[^"]|(?:(?<=\\)"))*
+EXIM_SUBJECT (T="%{EXIM_QUOTED_CONTENT:[exim][log][message][subject]}")
 
```
do not like introducing another pattern here but the alternative is even uglies using `(?<[exim][log]...`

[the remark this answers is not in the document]
yaauie: The lookbehind approach can also get tricked by an escaped backslash at the end (`T="banana\\"`).

What about:

```suggestion
EXIM_QUOTED_CONTENT (?:\\.|[^\\"])*
EXIM_SUBJECT (T="%{EXIM_QUOTED_CONTENT:[exim][log][message][subject]}")
```

Above, `EXIM_QUOTED_CONTENT` captures a sequence of zero or more:
 - backslash-anything two-char sequence; OR
 - character that is not a backslash or double-quote

If that doesn't work, I'm okay with remaining with `GREEDYDATA`.

yaauie (commented):
One remaining path forward for the quoted subject, but LGTM with either my suggestion or the `GREEDYDATA` necessary evil :)

PR #291, opened 2020-11-18
Feat: allow UNIXPATH to match non-ascii chars

kares:
`WINPATH` matches paths containin non-ascii chars but `UNIXPATH` does not, which makes `PATH` inconsistent.

e.g. `/var/vtaky/ďateľ/čierny` is a valid path name on Linux and should be matched ... 

The issue was discovered while bumping into #159 a longer path had chars with diacritics towards the end (such as `/home/eaeaea/data/import/Sample_xyz_lé_é_c b` from #292), this not only did not match but actually lead to a slow ingestion process as all of these entries needed to abort the extensive back-tracking with a grok timeout.

(follow-up on https://github.com/logstash-plugins/logstash-patterns-core/pull/292)

### DIFF
--- a/spec/patterns/core_spec.rb
+++ b/spec/patterns/core_spec.rb
@@ -236,7 +236,9 @@ describe "UNIXPATH" do
   end
 
   it "matches paths with non-ascii characters" do
-    expect(grok_match(pattern, '/opt/Čierný_Peter/.中', true)).to pass
+    event = build_event path = '/opt/Čierný_Peter/.中'
+    build_grok('UNIXPATH:path').filter event
+    expect( event.get('path') ).to eql path
   end
 
 end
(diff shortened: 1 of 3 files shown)


spec/patterns/core_spec.rb
```diff
@@ -234,6 +234,11 @@
       expect( event.to_hash['tags'] ).to include '_grokparsefailure'
     end
   end
+
+  it "matches paths with non-ascii characters" do
+    expect(grok_match(pattern, '/opt/Čierný_Peter/.中', true)).to pass
```
yaauie, line 239: I'm not terribly confident in the `to pass` bit, especially as our `pattern` appears to be unanchored. Can we favor using named captures and `to include(`...`)` instead to set firmer expectations about exactly what is matched/captured?

yaauie (commented):
+1 to the change and consistency.

I've left a nitpick about being more specific in what we specify in specs.

spec/patterns/core_spec.rb
kares: actually is anchored, the last `exact_match = true` [causes anchoring](https://github.com/logstash-plugins/logstash-patterns-core/blob/0204239/spec/spec_helper.rb#L36).

agree it's probably not super clear but this is what we've started with ...

PR #292, opened 2020-11-25
Fix: UNIXPATH to avoid DoS on long paths with unmatching chars

kares:
Using `PATH` and namely it's sub-pattern `UNIXPATH` can lead to DoS (due regular expression back-tracking).

Sample one line reproducer (pattern is the original expanded `UNIXPATH`):
`ruby -e '/(\/([\w_%!$@:.,+~-]+|\\\\.)*)+ /.match "/opt/abcdef/1/.22/3:3+3/foo@BAR/X-Y+Z/~Sample_l_&^_c b"'`

... the important part is the non-matching `&^` sequence at the end.

Currently `UNIXPATH` does not handle unicode chars (separate issue), this issue (100% CPU) was discovered with non-ascii chars at the end: `/home/eaeaea/data/import/Sample_xyz_lé_é_c b`

~~Also to be noted that the pattern does not interrupt properly, using a timeout, thus there isn't a good work-around except not using or redefining `PATH/UNIXPATH`.~~

resolves https://github.com/logstash-plugins/logstash-patterns-core/issues/159

### DIFF
--- a/patterns/grok-patterns
+++ b/patterns/grok-patterns
@@ -34,7 +34,7 @@ HOSTPORT %{IPORHOST}:%{POSINT}
 
 # paths
 PATH (?:%{UNIXPATH}|%{WINPATH})
-UNIXPATH (/([\w_%!$@:.,+~-]+|\\.)*)+
+UNIXPATH (/[\w_%!$@:.,+~-]+)+
 TTY (?:/dev/(pts|tty([pq])?)(\w+)?/?(?:[0-9]+))
 WINPATH (?>[A-Za-z]+:|\\)(?:\\[^\\?*]*)+
 URIPROTO [A-Za-z]([A-Za-z0-9+\-.]+)+
--- a/patterns/grok-patterns
+++ b/patterns/grok-patterns
@@ -32,7 +32,7 @@ HOSTNAME \b(?:[0-9A-Za-z][0-9A-Za-z-]{0,62})(?:\.(?:[0-9A-Za-z][0-9A-Za-z-]{0,62
 IPORHOST (?:%{IP}|%{HOSTNAME})
 HOSTPORT %{IPORHOST}:%{POSINT}
 
-# paths
+# paths (only absolute paths are matched)
 PATH (?:%{UNIXPATH}|%{WINPATH})
 UNIXPATH (/[\w_%!$@:.,+~-]*)+
 TTY (?:/dev/(pts|tty([pq])?)(\w+)?/?(?:[0-9]+))
--- a/patterns/grok-patterns
+++ b/patterns/grok-patterns
@@ -34,7 +34,7 @@ HOSTPORT %{IPORHOST}:%{POSINT}
 
 # paths
 PATH (?:%{UNIXPATH}|%{WINPATH})
-UNIXPATH (/[\w_%!$@:.,+~-]+)+
+UNIXPATH (/[\w_%!$@:.,+~-]*)+
 TTY (?:/dev/(pts|tty([pq])?)(\w+)?/?(?:[0-9]+))
 WINPATH (?>[A-Za-z]+:|\\)(?:\\[^\\?*]*)+
 URIPROTO [A-Za-z]([A-Za-z0-9+\-.]+)+
--- a/spec/patterns/core_spec.rb
+++ b/spec/patterns/core_spec.rb
@@ -118,7 +118,7 @@ describe "UNIXPATH" do
   context "dotted path" do
 
     it "should match path containing ." do
-      expect(grok_match(pattern, '/some/./path')).to pass
+      expect(grok_match(pattern, '/some/./path/')).to pass
       expect(grok_match(pattern, '/some/../path')).to pass
       expect(grok_match(pattern, '/../.')).to pass
       expect(grok_match(pattern, '/.')).to pass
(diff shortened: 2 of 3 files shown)


patterns/grok-patterns
```diff
@@ -34,7 +34,7 @@ HOSTPORT %{IPORHOST}:%{POSINT}
 
 # paths
 PATH (?:%{UNIXPATH}|%{WINPATH})
-UNIXPATH (/([\w_%!$@:.,+~-]+|\\.)*)+
+UNIXPATH (/[\w_%!$@:.,+~-]+)+
```
kares, line 37: really wanted to avoid the 3-level nested loop but I am certainly no regexp engine expert.

there's been a [separate attempt](https://github.com/logstash-plugins/logstash-patterns-core/pull/166) using atomic grouping. please let me know if the change seems to be breaking compatibility so we reconsider the alternative here.

spec/patterns/core_spec.rb
```diff
@@ -114,6 +114,83 @@
...
+  end
+
+  context "separators" do
+
+    it "should match root" do
+      expect(grok_match(pattern, '/')).to pass
+    end
+
+    it "should match" do # NOTE: we did not match these in < 4.2.0
+      expect(grok_match(pattern, '//')).to pass
```
kares, line 161: multiple slashes `//` did not match using the old pattern so this could be considered "breaking" but I would incline towards it rather being a bug as `//` is allowed according to the Unix path-name specification. 
there's a [good summary](https://unix.stackexchange.com/questions/256497/on-what-systems-is-foo-bar-different-from-foo-bar) of (mostly legacy) OSs these are not valid filesystem paths, for the curious.

patterns/grok-patterns
```diff
@@ -34,7 +34,7 @@ HOSTPORT %{IPORHOST}:%{POSINT}
 
 # paths
 PATH (?:%{UNIXPATH}|%{WINPATH})
-UNIXPATH (/([\w_%!$@:.,+~-]+|\\.)*)+
+UNIXPATH (/[\w_%!$@:.,+~-]*)+
```
kares, line 37: really wanted to avoid the 3-level nested loop but I am certainly no way near a regexp engine expert.

there's been a [separate attempt](https://github.com/logstash-plugins/logstash-patterns-core/pull/166) using atomic grouping. please let me know if the change seems to be "breaking" compatibility so we reconsider the alternative here.

spec/patterns/core_spec.rb
```diff
@@ -108,10 +108,99 @@
...
+      grok.register
+      grok
+    end
+
     let(:value) { 'a=/some/path, b=/some/other/path' }
 
+    it "should match both paths" do # but does not
+      event = build_event(value)
+      grok.filter(event)
+      expect( event.to_hash['a'] ).to eql '/some/path,'
```
kares, line 124: this case is an attempt to reproduce https://github.com/logstash-plugins/logstash-patterns-core/issues/13 which was supposed to be fixed using the [UNIXPATH](https://github.com/logstash-plugins/logstash-patterns-core/commit/06de0c1bc566874ff87d9b23765f4fee1b3a573f) (changed here) 
turns out these case weren't being handled (previous spec attempt gave a false impression) so the behaviour did not change in this PR.

spec/patterns/core_spec.rb
```diff
@@ -108,10 +108,99 @@
...
+    it "should match path starting with ." do
+      expect(grok_match(pattern, '../0')).to pass
+      expect(grok_match(pattern, './~')).to pass
+      expect(grok_match(pattern, '.../-')).to pass
+      expect(grok_match(pattern, './')).to pass
+      expect(grok_match(pattern, './,')).to pass
+      expect(grok_match(pattern, '../')).to pass
+      expect(grok_match(pattern, '.a/')).to pass
+      expect(grok_match(pattern, '.~/')).to pass
+    end
```
jsvd, line 154: we should beware that these don't actually pass. the matcher will try to match `${UNIXPATH}` to any part of that string. For example in `./~`, the test passes since the match is `/~`.
When tested in the context of a full string, we can see this failing:

```
❯ bin/logstash -e 'filter { grok { match => { "message" => [ "a %{UNIXPATH:path} b" ] }}}'               
a ./~ b
{
          "host" => "joaos-mbp.lan",
          "type" => "stdin",
      "@version" => "1",
    "@timestamp" => 2020-11-30T09:57:26.213Z,
          "tags" => [
        [0] "_grokparsefailure"
    ],
       "message" => "a ./~ b"
}
```

kares: thanks, updated specs to do exact matches and also spec-ed out these with some partial matching ...

spec/patterns/core_spec.rb
```diff
@@ -103,15 +103,122 @@
...
+      expect(grok_match(pattern, '.~/', true)).to_not pass
+    end
+
+    it "should not match if there's no separator" do
+      expect(grok_match(pattern, '.')).to_not pass
+      expect(grok_match(pattern, '..')).to_not pass
+      expect(grok_match(pattern, '...')).to_not pass
+      expect(grok_match(pattern, '.,')).to_not pass
+      expect(grok_match(pattern, '.-')).to_not pass
+    end
```
jsvd, line 180: we’re now officially stating that relative paths don’t work. this seems to be the same behaviour of the WINPATH (only absolute paths)
Can we wrap these two blocks with a `context "when path is relative"`, and add a comment in the patterns themselves saying that only absolute paths are matched?

jsvd (approved):
Overall LGTM, just added a note to comment in the patterns that they only support absolute paths.
Please squash this into 1 commit and mention that the change fixes certain ReDoS and also removes an attempt at supporting relative paths.

And also please open an issue for any remaining concerns, such as making UNIXPATH support non ascii characters.

PR #293, opened 2020-12-01
Feat: make firewall patterns ECS compliant (1/2)

kares:
This is the first part of converting the patterns (the other part will be reviewing all of our CISCO messages).

We used the opportunity to share some common ground between iptables based firewall backends.
Thus there's a new `IPTABLES` sub-pattern used by both `SHOREWALL` and `SFW2` (SuSE firewall 2).
Should also make the patterns more robust compared to our legacy matching (e.g. only IPv4 was matched previously for `SHOREWALL`).

There's a (minor) blocking [issue on grok](https://github.com/logstash-plugins/logstash-filter-grok/issues/165) which causes IPv6 fields (e.g. `iptables.length`) to not type cast properly.

Beats has a standalone [iptables](https://www.elastic.co/guide/en/beats/filebeat/master/exported-fields-iptables.html), we're using the same field naming conventions here.

Since `SFW2` and `SHOREWALL` both share most of their heavy lifting through the `IPTABLES` sub-pattern it should be enough to only review the one. Here's a sample use-case:

`May 28 17:23:25 myHost kernel: [3124658.791874] Shorewall:FORWARD:REJECT:IN=eth2 OUT=eth2 SRC=1.2.3.4 DST=192.168.0.10 LEN=141 TOS=0x00 PREC=0x00 TTL=63 ID=55251 PROTO=UDP SPT=5353 DPT=5335 LEN=121`

```
    "timestamp"=>"May 28 17:23:25",
    "host"=>{"hostname"=>"myHost"},
    "shorewall"=>{
        "firewall"=>{"type"=>"FORWARD", "action"=>"REJECT"}
    },
    "source"=>{"port"=>5353, "ip"=>"1.2.3.4"},
    "destination"=>{"port"=>5335, "ip"=>"192.168.0.10"}
    "network"=>{"transport"=>"UDP"},
    "iptables"=>{
        "input_interface"=>"eth2",
        "output_interface"=>"eth2",
        "length"=>141,
        "tos"=>"00",
        "precedence_bits"=>"00",
        "ttl"=>63,
        "id"=>"55251"
    }
```


spec/patterns/firewalls_spec.rb
```diff
@@ -2,135 +2,254 @@
...
-      expect(subject['src_ip']).to eq('192.168.1.1')
-      expect(subject['dst_ip']).to eq('www.example.com')
-      expect(subject['policy_id']).to eq('S_OUTSIDE_TO_INSIDE')
+  let(:message) do
+    "Jan 29 00:00:28 myth kernel: SFW2-INext-DROP-DEFLT IN=ppp0 OUT= MAC= SRC=24.64.208.134 DST=216.58.112.55 LEN=512 TOS=0x00 PREC=0x00 TTL=70 ID=55012 PROTO=UDP SPT=24128 DPT=1026 LEN=492"
+  end
+
+  it 'matches' do
+    # NOTE: we do not match the second LEN=492 which is the length of the wrapped (UDP in this case) packet
+    # iptables.length (IP packet length) 512 = 492 (UDP/TCP packet length) + 20 (IPv4 header length = 20 bytes)
```
kares, line 122: as noted here and also visible on the PR's description example we only capture the first `LEN=...` field as
`iptables.length` (IP packet length), second `LEN` represent (UDP/TCP packet length)`iptables.length - 20` (where `20` stands for IPv4 header length in bytes). does it make sense to capture the other as well, if so I was a bit lost on naming (`iptables.payload_length`) ?

kares:
// cc @ebeahan @webmat if you guys find some time double-checking (mostly the new) `IPTABLES` captures

patterns/ecs-v1/firewalls
```diff
@@ -83,9 +83,20 @@ CISCOFW713172 Group = %{GREEDYDATA:group}, IP = %{IP:src_ip}, Automatic NAT Dete
...
+
+IPTABLES IN=(?:%{NOTSPACE:[iptables][input_interface]})?\s+OUT=(?:%{NOTSPACE:[iptables][output_interface]})?\s+(?:MAC=(?:%{COMMONMAC:[destination][mac]})?(?::%{COMMONMAC:[source][mac]})?(?::[A-Fa-f0-9]{2}:[A-Fa-f0-9]{2})?\s+)?(:?%{IPTABLES4_PART}|%{IPTABLES6_PART}).*?PROTO=(?:%{WORD:[network][transport]})?\s+SPT=(?:%{INT:[source][port]:int})?\s+DPT=(?:%{INT:[destination][port]:int})?\s+(?:%{IPTABLES_TCP_PART})?
+
 # Shorewall firewall logs
-SHOREWALL (%{SYSLOGTIMESTAMP:timestamp}) (%{WORD:nf_host}) kernel:.*Shorewall:(%{WORD:nf_action1})?:(%{WORD:nf_action2})?.*IN=(%{USERNAME:nf_in_interface})?.*(OUT= *MAC=(%{COMMONMAC:nf_dst_mac}):(%{COMMONMAC:nf_src_mac})?|OUT=%{USERNAME:nf_out_interface}).*SRC=(%{IPV4:nf_src_ip}).*DST=(%{IPV4:nf_dst_ip}).*LEN=(%{WORD:nf_len}).?*TOS=(%{WORD:nf_tos}).?*PREC=(%{WORD:nf_prec}).?*TTL=(%{INT:nf_ttl}).?*ID=(%{INT:nf_id}).?*PROTO=(%{WORD:nf_protocol}).?*SPT=(%{INT:nf_src_port}?.*DPT=%{INT:nf_dst_port}?.*)
+SHOREWALL (?:%{SYSLOGTIMESTAMP:timestamp}) (?:%{WORD:[host][hostname]}) .*Shorewall:(?:%{WORD:[shorewall][firewall][type]})?:(?:%{WORD:[shorewall][firewall][action]})?.*%{IPTABLES}
```
webmat, line 97: Instead of populating `[host][hostname]` here I'd recommend populating `[observer][hostname]` instead.

Observers are any monitoring agent that monitors from the outside, like firewalls.

patterns/ecs-v1/firewalls
```diff
@@ -83,9 +83,20 @@ CISCOFW713172 Group = %{GREEDYDATA:group}, IP = %{IP:src_ip}, Automatic NAT Dete
...
 #== End Shorewall
 #== SuSE Firewall 2 ==
-SFW2 ((%{SYSLOGTIMESTAMP})|(%{TIMESTAMP_ISO8601}))\s*%{HOSTNAME}\s*kernel\S+\s*%{NAGIOSTIME}\s*SFW2\-INext\-%{NOTSPACE:nf_action}\s*IN=%{USERNAME:nf_in_interface}.*OUT=((\s*%{USERNAME:nf_out_interface})|(\s*))MAC=((%{COMMONMAC:nf_dst_mac}:%{COMMONMAC:nf_src_mac})|(\s*)).*SRC=%{IP:nf_src_ip}\s*DST=%{IP:nf_dst_ip}.*PROTO=%{WORD:nf_protocol}((.*SPT=%{INT:nf_src_port}.*DPT=%{INT:nf_dst_port}.*)|())
+SFW2_LOG_PREFIX SFW2\-INext\-%{NOTSPACE:[suse][firewall][action]}
+SFW2 ((?:%{SYSLOGTIMESTAMP:timestamp})|(?:%{TIMESTAMP_ISO8601:timestamp}))\s*%{HOSTNAME:[host][hostname]}.*?%{SFW2_LOG_PREFIX:[suse][firewall][log_prefix]}\s*%{IPTABLES}
```
webmat, line 101: `[observer][hostname]` here as well.

webmat (commented):
Looking good overall! I do have a few recommendations about using the "observer" fields below.

I don't know much about the corresponding Beats modules. +1 on matching what they're doing, but you might benefit from pinging one of the folks who've worked on those as well to get their take :-)

patterns/ecs-v1/firewalls
kares: Thanks Mat, these felt weird indeed. Updated!

kares:
@yaauie this is now good for a LS review - there's minor legacy updates (non-breaking) as I was lazy to write more specs to work-around the issue of the `SFW2` being broken when `OUT=eth0` is specified in legacy mode.

spec/patterns/shorewall_spec.rb
```diff
@@ -2,89 +2,101 @@
...
-    it { should include("nf_src_port" => "59088") }
+    let(:message) do
+      "May 28 17:31:07 server Shorewall:net2fw:DROP:" +
+      "IN=eth1 OUT= MAC=00:02:b3:c7:2f:77:38:72:c0:6e:92:9c:08:00 SRC=127.0.0.1 DST=1.2.3.4 LEN=60 TOS=0x00 PREC=0x00 TTL=49 ID=6480 DF PROTO=TCP SPT=59088 DPT=8080 WINDOW=2920 RES=0x00 SYN URGP=0"
+    end
+
+    it 'matches' do
+      should include("timestamp" => "May 28 17:31:07")
+      if ecs_compatibility?
+        expect(grok).to include(
```
yaauie, line 63: nit: using [rspec's `aggregate_failures`](https://relishapp.com/rspec/rspec-expectations/docs/aggregating-failures) can help tell the whole story a lot better in cases where we have many unchained expectations about the result of a single action.

spec/patterns/shorewall_spec.rb
```diff
@@ -2,89 +2,101 @@
...
-    it { should include("nf_protocol" => "TCP") }
 
-    it { should include("nf_src_port" => "59088") }
+    let(:message) do
+      "May 28 17:31:07 server Shorewall:net2fw:DROP:" +
+      "IN=eth1 OUT= MAC=00:02:b3:c7:2f:77:38:72:c0:6e:92:9c:08:00 SRC=127.0.0.1 DST=1.2.3.4 LEN=60 TOS=0x00 PREC=0x00 TTL=49 ID=6480 DF PROTO=TCP SPT=59088 DPT=8080 WINDOW=2920 RES=0x00 SYN URGP=0"
+    end
+
+    it 'matches' do
+      should include("timestamp" => "May 28 17:31:07")
```
yaauie, line 61: nit: I really don't like the `should` short-hand with implicit subject. While it makes the `it { should ` ... ` }` one-liners more "readable", I think it does the opposite for long-form blocks.

I would much prefer an explicit subject:

~~~ ruby
        subject(:the_event) { grok_match(pattern, value) }
~~~

And the `expect` syntax:

~~~
  it 'matches' do
    expect(the_event).to include("timestamp" => "May 28 17:31:07")
    # ...
  end
~~~

yaauie (commented):
inasmuch as I understand the context, the changes to the patterns and the naming/structure make a lot of sense to me. The specs certainly tell a good story and make it easy to be confident that the patterns work as intended.

I've left a couple nitpicks about readability of the specs in the comments, but other than that LGTM. 

Fantastic work.

spec/patterns/shorewall_spec.rb
kares: yeah, :+1: also hate the should syntax.
these specs did get boring so I ended up :scissors: based on legacy, instead of rewriting everything to expect.

spec/patterns/shorewall_spec.rb
kares: :bow: interesting feature, not really an rspec person I just suffer my way through specs :wink: 
in this particular case I do like the failures one-by-one since there likely will be multiple cases failing due the same cause already, as the whole file is run. these specs could have been much better have I spent more time writing a few helpers...

thanks for the hint again, will consider using this feature.

spec/patterns/shorewall_spec.rb
kares: might address some of these refactorings in the second part, so that we can get a review for the whole pattern set.

PR #294, opened 2020-12-21
Feat: ECS compliant Juniper (SRX) RT_FLOW captures

kares:
These should be fairly straight-forward - Beats has a [Juniper module](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-module-juniper.html).

There were no existing specs but according to Juniper KB these match unstructured SRX gateway logs.

NOTE: Beats only supports [structured logs](https://github.com/elastic/beats/blob/v7.10.1/x-pack/filebeat/module/juniper/srx/test/flow.log#L1) while here LS is aimed at parsing unstructured log lines.
For better interoperability we follow the `juniper.srx` prefix naming for custom fields established by Beats.

### DIFF
--- a/patterns/ecs-v1/junos
+++ b/patterns/ecs-v1/junos
@@ -1,9 +1,11 @@
 # JUNOS 11.4 RT_FLOW patterns
-RT_FLOW_EVENT (RT_FLOW_SESSION_CREATE|RT_FLOW_SESSION_CLOSE|RT_FLOW_SESSION_DENY)
+RT_FLOW_TAG (?:RT_FLOW_SESSION_CREATE|RT_FLOW_SESSION_CLOSE|RT_FLOW_SESSION_DENY)
+# deprecated legacy name:
+RT_FLOW_EVENT RT_FLOW_TAG
 
-RT_FLOW1 %{RT_FLOW_EVENT:event}: %{GREEDYDATA:close-reason}: %{IP:src-ip}/%{INT:src-port}->%{IP:dst-ip}/%{INT:dst-port} %{DATA:service} %{IP:nat-src-ip}/%{INT:nat-src-port}->%{IP:nat-dst-ip}/%{INT:nat-dst-port} %{DATA:src-nat-rule-name} %{DATA:dst-nat-rule-name} %{INT:protocol-id} %{DATA:policy-name} %{DATA:from-zone} %{DATA:to-zone} %{INT:session-id} \d+\(%{DATA:sent}\) \d+\(%{DATA:received}\) %{INT:elapsed-time} .*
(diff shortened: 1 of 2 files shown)


patterns/ecs-v1/junos
```diff
@@ -1,9 +1,11 @@
...
+RT_FLOW_EVENT RT_FLOW_TAG
 
-RT_FLOW1 %{RT_FLOW_EVENT:event}: %{GREEDYDATA:close-reason}: %{IP:src-ip}/%{INT:src-port}->%{IP:dst-ip}/%{INT:dst-port} %{DATA:service} %{IP:nat-src-ip}/%{INT:nat-src-port}->%{IP:nat-dst-ip}/%{INT:nat-dst-port} %{DATA:src-nat-rule-name} %{DATA:dst-nat-rule-name} %{INT:protocol-id} %{DATA:policy-name} %{DATA:from-zone} %{DATA:to-zone} %{INT:session-id} \d+\(%{DATA:sent}\) \d+\(%{DATA:received}\) %{INT:elapsed-time} .*
+RT_FLOW1 %{RT_FLOW_TAG:[juniper][srx][tag]}: %{GREEDYDATA:[juniper][srx][reason]}: %{IP:[source][ip]}/%{INT:[source][port]:int}->%{IP:[destination][ip]}/%{INT:[destination][port]:int} %{DATA:[juniper][srx][service_name]} %{IP:[source][nat][ip]}/%{INT:[source][nat][port]:int}->%{IP:[destination][nat][ip]}/%{INT:[destination][nat][port]:int} (?:(?:None)|(?:%{DATA:[juniper][srx][src_nat_rule_name]})) (?:(?:None)|(?:%{DATA:[juniper][srx][dst_nat_rule_name]})) %{INT:[network][iana_number]} %{DATA:[rule][name]} %{DATA:[observer][ingress][zone]} %{DATA:[observer][egress][zone]} %{INT:[juniper][srx][session_id]} \d+\(%{DATA:[source][bytes]:int}\) \d+\(%{DATA:[destination][bytes]:int}\) %{INT:[juniper][srx][elapsed_time]:int} .*
```
kares, line 6: should `juniper.srx.reason` be `event.reason` (we might use `event.reason` in some firewall captures) ?

patterns/ecs-v1/junos
```diff
@@ -1,9 +1,11 @@
...
 
-RT_FLOW2 %{RT_FLOW_EVENT:event}: session created %{IP:src-ip}/%{INT:src-port}->%{IP:dst-ip}/%{INT:dst-port} %{DATA:service} %{IP:nat-src-ip}/%{INT:nat-src-port}->%{IP:nat-dst-ip}/%{INT:nat-dst-port} %{DATA:src-nat-rule-name} %{DATA:dst-nat-rule-name} %{INT:protocol-id} %{DATA:policy-name} %{DATA:from-zone} %{DATA:to-zone} %{INT:session-id} .*
+RT_FLOW2 %{RT_FLOW_TAG:[juniper][srx][tag]}: session created %{IP:[source][ip]}/%{INT:[source][port]:int}->%{IP:[destination][ip]}/%{INT:[destination][port]:int} %{DATA:[juniper][srx][service_name]} %{IP:[source][nat][ip]}/%{INT:[source][nat][port]:int}->%{IP:[destination][nat][ip]}/%{INT:[destination][nat][port]:int} (?:(?:None)|(?:%{DATA:[juniper][srx][src_nat_rule_name]})) (?:(?:None)|(?:%{DATA:[juniper][srx][dst_nat_rule_name]})) %{INT:[network][iana_number]} %{DATA:[rule][name]} %{DATA:[observer][ingress][zone]} %{DATA:[observer][egress][zone]} %{INT:[juniper][srx][session_id]} .*
```
kares, line 8: `[juniper][srx][src_nat_rule_name]` looks a bit ugly but this is the field name convention of the original filename,
since there isn't an ECS counterpart.

kares:
// cc @ebeahan @webmat if you get a chance for a quick field naming review - specs [show-case](https://github.com/logstash-plugins/logstash-patterns-core/pull/294/files#diff-db5db76708ae1859692bad8dea2847bd72d7ab7930b9197eb4f698a270ef5caeR18-R26) the ecs captures

patterns/ecs-v1/junos
ebeahan: `event.reason` can pair with `event.action` capture the _why_ of an event - something like `event.action: blocked` paired with a reason of `blocked due to web filter`.

Here, `event.action` isn't populated and the reason just reiterates that the session has been closed. I don't think it's necessary to switch to `event.reason` here.

yaauie (approved):
🎉 LGTM

Issue #243, opened 2019-04-18
HTTPD_COMBINEDLOG pattern conflicts with ECS v1.0


C-Duv:
The `HTTPD_COMBINEDLOG` pattern creates a `agent` field which conflicts with [the `agent` field of ECS 1.0](https://github.com/elastic/ecs/blob/1.0/fields.yml#L8).

hume-github:
I'm running into this as well.  I'd propose renaming the match to "useragent", and I will in my own configuration, but I'm not sure what kind of downstream effects that'll have.

holiiveira:
any workaround for this?

tcassaert:
What I've done is replacing the HTTPD_COMBINEDLOG pattern with `%{HTTPD_COMMONLOG} %{QS:referrer} %{QS:client_agent}`. This avoids having a conflicting `agent` field.

danlsgiga:
Faced the same issue and used the workaround @tcassaert suggested. I'd suggest to have the `agent` changed to `useragent`

Issue #173, opened 2016-09-22
revisit existing patterns to add more type casting


jsvd:
This issue's purpose is to open the discussion to add more casting to existing patterns.

Take for example the following pattern:

```
patterns/firewalls:CISCOFW313005 %{CISCO_REASON:reason} for %{WORD:protocol} error message: %{WORD:err_protocol} src %{DATA:err_src_interface}:%{IP:err_src_ip}(\(%{DATA:err_src_fwuser}\))? dst %{DATA:err_dst_interface}:%{IP:err_dst_ip}(\(%{DATA:err_dst_fwuser}\))? \(type %{INT:err_icmp_type}, code %{INT:err_icmp_code}\) on %{DATA:interface} interface\.  Original IP payload: %{WORD:protocol} src %{IP:orig_src_ip}/%{INT:orig_src_port}(\(%{DATA:orig_src_fwuser}\))? dst %{IP:orig_dst_ip}/%{INT:orig_dst_port}(\(%{DATA:orig_dst_fwuser}\))?
```

There are multiple INT patterns which aren't cast to integer, which forces users to add mutate filters after applying the grok pattern.

I don't think we can simply force all INT patterns to be converted to ints and all FLOAT's to float, but we can audit existing composed patterns and add :int / :float to some.

For example, transform the pattern above into:

```
patterns/firewalls:CISCOFW313005 %{CISCO_REASON:reason} for %{WORD:protocol} error message: %{WORD:err_protocol} src %{DATA:err_src_interface}:%{IP:err_src_ip}(\(%{DATA:err_src_fwuser}\))? dst %{DATA:err_dst_interface}:%{IP:err_dst_ip}(\(%{DATA:err_dst_fwuser}\))? \(type %{INT:err_icmp_type:int}, code %{INT:err_icmp_code:int}\) on %{DATA:interface} interface\.  Original IP payload: %{WORD:protocol} src %{IP:orig_src_ip}/%{INT:orig_src_port:int}(\(%{DATA:orig_src_fwuser}\))? dst %{IP:orig_dst_ip}/%{INT:orig_dst_port:int}(\(%{DATA:orig_dst_fwuser}\))?
```

nellicus:
indeed a step forward @jsvd thanks for starting this.

it'd be great to have some way to have logstash attempt to recognize the type a-la elasticsearch however I ignore how doable that'd be

nellicus:
@jsvd 

[...]

PR #297, opened 2020-12-30
Feat: fully ECS compliant captures

### LINKED ISSUE #233: remove named capture groups from predefined grok patterns
Named capture groups create new fields in logstash and can cause unexpected behaviour. In most patterns that is not the case, but a handful of them name some fields.

For example, when using the `URIHOST` grok pattern, a field with the name `port` is created as one of the capture groups is named in the pattern definition. 
Changing this would be a breaking change as some users might be relying on this behaviour. We could fix this in time for 7.x 

`URIHOST %{IPORHOST}(?::%{POSINT:port})?`

should be changed to

`URIHOST %{IPORHOST}(?::%{POSINT})?` 

My suggestion is to remove all named capture groups from https://github.com/logstash-plugins/logstash-patterns-core/blob/master/patterns/grok-patterns 

Other patterns should continue to have name groups to make them easier to use, for example https://github.com/logstash-plugins/logstash-patterns-core/blob/master/patterns/haproxy

### LINKED ISSUE #248: Logstash plugins create events breaking ECS
Logstash 7.3
OS: All
Logstash direct input plugins such as tcp, udp, syslog, etc. create events with the field "host" populated incorrectly. To conform to ECS the host name should be placed into "host.name". It would be useful if the plugins could also populate other fields like "host.ip". 

Arguably it should populate "agent" fields as well since logstash is acting as the agent.

### LINKED ISSUE #258: Field agent collides with filebeat agent field in grok COMBINEDAPACHELOG pattern
As confirmed by [Pier-Hugues Pellerin](https://discuss.elastic.co/u/pierhugues) in [this discussion](https://discuss.elastic.co/t/logstash-grok-pattern-combinedapachelog-field-name-collision-with-latest-filebeat-7-2-0-client-and-logstash-useragent-plugin/191584), `agent` field in filebeat collides with `agent` field in `COMBINEDAPACHELOG` **grok** pattern:
```
COMBINEDAPACHELOG %{COMMONAPACHELOG} %{QS:referrer} %{QS:agent}
```

### Workaround
Expand `%{COMBINEDAPACHELOG}` and replace `agent` with something else:
```
%{COMMONAPACHELOG} %{QS:referrer} %{QS:somethingelse}
```

### Version afflicted 
- filebeat 7.2.X (from discussion)
- filebeat 7.6.1

### LINKED ISSUE #278: Implement ECS-Compatibility Mode
this one falls under: https://github.com/logstash-plugins/logstash-filter-grok/issues/157

There's going to be 2 set of patterns provided out-of-the-box that provide around the same functionality (at start).
The existing (legacy) set and an ECS set of patterns - with mostly captures reviewed/renamed for compliance.

Patterns are already logically split by functionality into separate files - good enough to convert the whole set one-by-one :
- [x] aws https://github.com/logstash-plugins/logstash-patterns-core/pull/287
- [x] bacula https://github.com/logstash-plugins/logstash-patterns-core/pull/295
- [x] bind https://github.com/logstash-plugins/logstash-patterns-core/pull/281
- [x] bro https://github.com/logstash-plugins/logstash-patterns-core/pull/286
- [x] exim https://github.com/logstash-plugins/logstash-patterns-core/pull/283 (base-line) #290
- [x] firewalls https://github.com/logstash-plugins/logstash-patterns-core/pull/293 and https://github.com/logstash-plugins/logstash-patterns-core/pull/296
- [x] grok-patterns - core set https://github.com/logstash-plugins/logstash-patterns-core/pull/262
- [x] haproxy https://github.com/logstash-plugins/logstash-patterns-core/pull/289
- [x] httpd #280 #267
- [x] java https://github.com/logstash-plugins/logstash-patterns-core/pull/268
- [x] junos https://github.com/logstash-plugins/logstash-patterns-core/pull/294
- [x] linux-syslog https://github.com/logstash-plugins/logstash-patterns-core/pull/262
- [x] ~~maven~~
- [x] mcollective+mcollective-patterns https://github.com/logstash-plugins/logstash-patterns-core/pull/271
- [x] mongodb https://github.com/logstash-plugins/logstash-patterns-core/pull/269
- [x] nagios https://github.com/logstash-plugins/logstash-patterns-core/pull/276
- [x] postgresql https://github.com/logstash-plugins/logstash-patterns-core/pull/266
- [x] rails https://github.com/logstash-plugins/logstash-patterns-core/pull/265
- [x] redis https://github.com/logstash-plugins/logstash-patterns-core/pull/263
- [x] ruby https://github.com/logstash-plugins/logstash-patterns-core/pull/265
- [x] squid https://github.com/logstash-plugins/logstash-patterns-core/pull/270

Post TODOs

- ship (legacy) pattern updates 4.2.0 before ECS-ified release

- check if ECS 1.6 is around with `http.response.mime_type` https://github.com/logstash-plugins/logstash-patterns-core/pull/270#issuecomment-670012568
  ~~in the mean time consider renaming `squid.response.content_type` to `squid.response.mime_type` ?~~
  UPDATE 1.7 shipped `http.[request|response].mime_type` as GA
  :heavy_check_mark: *renamed `squid.response.content_type` at https://github.com/logstash-plugins/logstash-patterns-core/commit/3b8655765c606ae3763ddfcdbbdaea701e82cfbf*
  
- `host.name` vs `host.hostname`
  https://github.com/logstash-plugins/logstash-patterns-core/pull/262/files#r459038278 
  https://github.com/logstash-plugins/logstash-patterns-core/pull/262/files#r504850250
  :heavy_check_mark: *we'll be using `hostname`, users should opt-in to `name` (due SIEM)*
- introduce grok conversions for ES ingest node compatibility (strict `:int` and `:float` parsing modes, add `:long`, `:double`, `:boolean` + we could also use an AS style `:bool` conversion e.g. with bro's `T`/`F` flags)
  than review patterns
  :heavy_check_mark: *won't block shipping ECS-ified patterns - we should double check :int capture INT-like values*
  :white_check_mark: `...:int` captures reviewed  https://github.com/logstash-plugins/logstash-patterns-core/commit/6dd657b8bff1604439d17e29c8b2d2a9bbd6dd12

- need to make sure to set `event.original` (in grok?) and potentially remove `message` field before reaching grok
  [patterns that extract `message`]( https://github.com/logstash-plugins/logstash-patterns-core/pull/276#discussion_r520585904) to keep the field flat (or default to running with `overwrite => [ 'message' ]`) in ECS mode

- type-casting doesn't always work due a grok library [bug](https://github.com/logstash-plugins/logstash-filter-grok/issues/165)
  
- ~~avoid mixed <style> named captures (https://github.com/logstash-plugins/logstash-patterns-core/pull/265#discussion_r466515254)~~

kares:
The joint effort from ECS-ification of pattern files (see meta-issue: #278).
Additionally also includes: #298, #299, #273

Commits outside individual file updates (NOT reviewed):
- https://github.com/logstash-plugins/logstash-patterns-core/pull/297/commits/6dd657b8bff1604439d17e29c8b2d2a9bbd6dd12 (review of :int/:float coercions)
- https://github.com/logstash-plugins/logstash-patterns-core/pull/297/commits/730ab4f6d64dac861c114dd1fb03dce331fd6c3b (bump+changelog)
- https://github.com/logstash-plugins/logstash-patterns-core/pull/297/commits/39907daab82ee8cedf1713cf4b405aa9de187369 (`:long` coercion 'annotations' after each pattern capturing an `:int` that might potentially overflow)

### CHANGELOG

- **aws**
  * in ECS mode we dropped the (incomplete) attempt to capture `rawrequest` from `S3_REQUEST_LINE`
  * `S3_ACCESS_LOG` will handle up-to-date S3 access-log formats (6 'new' field captures at the end)
    Host Id -> Signature Version -> Cipher Suite -> Authentication Type -> Host Header -> TLS version
  * `ELB_ACCESS_LOG` will handle optional (`-`) in legacy mode
  * null values such as `-` or `-1` time values (e.g. `ELB_ACCESS_LOG`'s `request_processing_time`)
    are not captured in ECS mode

- **bacula**
  - Fix: improve matching of `BACULA_HOST` as `HOSTNAME`
  - Fix: legacy `BACULA_` patterns to handle (optional) spaces
  - Fix: handle `BACULA_LOG` 'Job Id: X' prefix as optional
  - Fix: legacy matching of BACULA fatal error lines

- **bind**
  - `BIND9`'s legacy `querytype` was further split into multiple fields as:
     `dns.question.type` and `bind.log.question.flags`
  - `BIND9` patterns (legacy as well) were adjusted to handle Bind9 >= 9.11 compatibility
  - `BIND9_QUERYLOGBASE` was introduced for potential re-use

- **bro**
  * `BRO_` patterns are stricter in ECS mode - won't mistakenly match newer BRO/Zeek formats
  * place holders such as `(empty)` tags and `-` null values won't be captured
  * each `BRO_` pattern has a newer `ZEEK_` variant that supports latest Zeek 3.x versions
    e.g. `ZEEK_HTTP` as a replacement for `BRO_HTTP` (in ECS mode only),
    there's a new file **zeek** where all of the `ZEEK_XXX` pattern variants live

- **exim**
  * introduced `EXIM` (`EXIM_MESSAGE_ARRIVAL`) to match message arrival log lines - in ECS mode!

- **firewalls**
  * introduced `IPTABLES` pattern which is re-used within `SHOREWALL` and `SFW2`
  * `SHOREWALL` now supports IPv6 addresses (in ECS mode - due `IPTABLES` pattern)
  * `timestamp` fields will be captured for `SHOREWALL` and `SFW2` in legacy mode as well
  * `SHOREWALL` became less strict in containing the `kernel:` sub-string
  * `NETSCREENSESSIONLOG` properly handles optional `session_id=... reason=...` suffix
  * `interval` and `xlate_type` (legacy) CISCO fields are not captured in ECS mode

- **core** (grok-patterns)
  * `SYSLOGFACILITY` type casts facility code and priority in ECS mode
  * `SYSLOGTIMESTAMP` will be captured (from `SYSLOGBASE`) as `timestamp`
  * Fix: e-mail address's local part to match according to RFC (#273)

- **haproxy**
  * several ECS-ified fields will be type-casted to integer in ECS mode e.g. *haproxy.bytes_read*
  * fields containing null value (`-`) are no longer captured
    (e.g. in legacy mode `captured_request_cookie` gets captured even if `"-"`)

- **httpd**
  * optional fields (e.g. `http.request.referrer` or `user_agent`) are only captured when not null (`-`)
  * `source.port` (`clientport` in legacy mode) is considered optional
  * dropped raw data (`rawrequest` legacy field) in ECS mode
  * Fix: HTTPD_ERRORLOG should match when module missing (#299)

- **java**
  * `JAVASTACKTRACEPART`'s matched line number will be converted to an integer
  * `CATALINALOG` matching was updated to handle Tomcat 7/8/9 logging format
  * `TOMCATLOG` handles the default Tomcat 7/8/9 logging format
  * old (custom) legacy TOMCAT format is handled by the added `TOMCATLEGACY_LOG`
  * `TOMCATLOG` and `TOMCAT_DATESTAMP` still match the legacy format, 
      however this might change at a later point - if you rely on the old format use `TOMCATLEGACY_` patterns

- **junos**
  * integer fields (e.g. `juniper.srx.elapsed_time`) are captured as integer values

- **linux-syslog**
  * `SYSLOG5424LINE` captures (overwrites) the `message` field instead of using a custom field name
  * regardless of the format used, in ECS mode, timestamps are always captured as `timestamp`
  * fields such as `log.syslog.facility.code` and `process.pid` are converted to integers

- **mcollective**
  * *mcollective-patterns* file was removed, it's all one *mcollective* in ECS mode
  * `MCOLLECTIVE`'s `process.pid` (`pid` previously) is not type-casted to an integer

- **nagios**
  * numeric fields such as `nagios.log.attempt` are converted to integer values in ECS mode

- **rails**
  * request duration times from `RAILS3` log will be converted to floating point values

- **squid**
  * `SQUID3`'s `duration` http.response `status_code` and `bytes` are type-casted to int
  * `SQUID3` pattern won't capture null ('-') `user.name` or `squid.response.content_type`
  * Fix: allow to parse SQUID log with status 0 (#298)
  * Fix: handle optional server address (#298)

---
resolves #278
fixes #248 fixes #258 closing #243 fixes #233 closing #173


Gemfile
```diff
@@ -9,3 +9,6 @@ if Dir.exist?(logstash_path) && use_logstash_source
   gem 'logstash-core', :path => "#{logstash_path}/logstash-core"
   gem 'logstash-core-plugin-api', :path => "#{logstash_path}/logstash-core-plugin-api"
 end
+
+# TODO till filter grok with ECS support is released :
+gem 'logstash-filter-grok', git: 'https://github.com/kares/logstash-filter-grok.git', ref: 'ecs-1-support'
```
kares, line 14: for a :green_circle: CI this needs to be in until the filter (depending on this plugin) is also released.

patterns/ecs-v1/grok-patterns
```diff
@@ -0,0 +1,96 @@
...
+# paths (only absolute paths are matched)
+PATH (?:%{UNIXPATH}|%{WINPATH})
+UNIXPATH (/[[[:alnum:]]_%!$@:.,+~-]*)+
+TTY (?:/dev/(pts|tty([pq])?)(\w+)?/?(?:[0-9]+))
+WINPATH (?>[A-Za-z]+:|\\)(?:\\[^\\?*]*)+
+URIPROTO [A-Za-z]([A-Za-z0-9+\-.]+)+
+URIHOST %{IPORHOST}(?::%{POSINT})?
+# uripath comes loosely from RFC1738, but mostly from what Firefox doesn't turn into %XX
+URIPATH (?:/[A-Za-z0-9$.+!*'(){},~:;=@#%&_\-]*)+
+#URIQUERY (?:[A-Za-z0-9]+(?:=(?:[^&]*))?(?:&(?:[A-Za-z0-9]+(?:=(?:[^&]*))?)?)*)?
```
jsvd, line 44: can this be removed?

patterns/legacy/exim
```diff
@@ -1,13 +1,19 @@
...
 EXIM_INTERFACE (I=\[%{IP:exim_interface}\](:%{NUMBER:exim_interface_port}))
 EXIM_PROTOCOL (P=%{NOTSPACE:protocol})
 EXIM_MSG_SIZE (S=%{NUMBER:exim_msg_size})
 EXIM_HEADER_ID (id=%{NOTSPACE:exim_header_id})
 EXIM_SUBJECT (T=%{QS:exim_subject})
 
+EXIM_UNKNOWN_FIELD (?:[A-Za-z0-9]{1,4}=%{NOTSPACE})
+EXIM_NAMED_FIELDS (?: (?:%{EXIM_REMOTE_HOST}|%{EXIM_INTERFACE}|%{EXIM_PROTOCOL}|%{EXIM_MSG_SIZE}|%{EXIM_HEADER_ID}|%{EXIM_SUBJECT}|%{EXIM_UNKNOWN_FIELD}))*
+
+EXIM_MESSAGE_ARRIVAL %{EXIM_DATE:timestamp} (?:%{EXIM_PID} )?%{EXIM_MSGID:exim_msgid} (?<exim_flags><=) (?<exim_status>[a-z:] )?%{EMAILADDRESS:exim_sender_email}%{EXIM_NAMED_FIELDS}(?: for %{EMAILADDRESS:exim_recipient_email})?
```
jsvd, line 17: Could we perhaps move the `:pid` field assigment from EXIM_PID to the place where it's used, here:

```suggestion
EXIM_MESSAGE_ARRIVAL %{EXIM_DATE:timestamp} (?:%{EXIM_PID:pid} )?%{EXIM_MSGID:exim_msgid} (?<exim_flags><=) (?<exim_status>[a-z:] )?%{EMAILADDRESS:exim_sender_email}%{EXIM_NAMED_FIELDS}(?: for %{EMAILADDRESS:exim_recipient_email})?
```

Skimming through https://github.com/logstash-plugins/logstash-patterns-core/pull/283 I couldn't find a reason for the change

jsvd (commented):
All and all this a TREMENDOUS effort, awesome job pulling this together 🎉 
The commends are left are mostly nitpicks, as @yaauie and @webmat already did a thorough review from the grok and ECS perspectives.

.ci/setup.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -ex
+
+sudo yum install -y git
```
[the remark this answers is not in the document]
kares: we will need this until [grok filter is released](https://github.com/logstash-plugins/logstash-filter-grok/pull/162) - since they are interdependent on each other

patterns/ecs-v1/grok-patterns
kares: this seems like a left-over but it's actually what exists on the legacy end (just that it got a rename):
https://github.com/logstash-plugins/logstash-patterns-core/blob/v4.2.0/patterns/grok-patterns#L45

patterns/legacy/exim
kares: it's due to `EXIM_PID \[%{POSINT:pid}\]` so we would capture the `[12345]` instead of `12345`

patterns/bind
```diff
@@ -1,3 +0,0 @@
-BIND9_TIMESTAMP %{MONTHDAY}[-]%{MONTH}[-]%{YEAR} %{TIME}
```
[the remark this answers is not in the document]
kares: hmm, this is weird - it was intentionally done as a rename for all: https://github.com/logstash-plugins/logstash-patterns-core/pull/297/commits/f7977c79566eaf071e9f396a085320254b836001?file-filters%5B%5D=No+extension
not sure what's going on, maybe the master merges messed with the rename ... on it, :bow:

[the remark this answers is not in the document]
kares: turns out this is how git diff rename detection works. 
if we squash and merge we'll loose history as outlined by the diff - as renames are detected based on similarity and some of the renamed files changed in ways that it's not being detected. 

especially tricky as the same files are than also re-added under a different dir (e.g. https://github.com/logstash-plugins/logstash-patterns-core/compare/master...deb476). 

one option might be to try squashing into fewer logical commits - let's see if it gets tricky with upstream merges.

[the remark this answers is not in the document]
kares: a good compromise seems to be to keep at least the commit with the `patterns/*` -> `patterns/legacy/*` rename 
in a standalone commit and then we can squash the rest into one commit (end result is 2 commits) - the diff algorithm won't get confused on those small files where changing a line is considered changing 50% of file content by git.

PR #215, opened 2017-12-11
Minor update for some cases I found in my logs

lingfish:

PR #298, opened 2021-01-05
Fix: allow to parse SQUID log with status 0 + handle optional server address

kares:
closing #215


yaauie (commented):
- looks to address the `000` status code on abort from #215, but not the `-` value for its `server` capture -- if this is intentional, could you add a note as to why?
 - 👍🏼 to not capturing the status code in ECS.

kares:
thanks - good catch, this is the thing with PRs that do not provide tests/details - its easy to miss all the changes ...

decided to handle the `000` status (had a sample log message collected previosly) and I've seen a PR around.,
thought it's the same issue. for the optional `-` server capture I will need to look into squid and ideally add a spec.

PR #302, opened 2021-02-17
Build: revert the need to use a git source

kares:
... after https://github.com/logstash-plugins/logstash-filter-grok/pull/162 (logstash-filter-grok is released)


andsel (approved):
LGTM

PR #303, opened 2021-03-15
Fix: incorrect syslog (priority) field name

kares:
a follow-up on https://github.com/logstash-plugins/logstash-patterns-core/pull/297

- for `SYSLOG5424PRI` we clearly want `log.syslog.priority`
- same for `CISCO_TAGGED_SYSLOG`
  this field slipped through ECS-ification (no proper use-case with the specific pattern files e.g. `CISCOFW104001`)
  we clearly missed ECS-ifing the `ciscotag` field name (as `cisco.asa.tag`)


kaisecheng (approved):
LGTM 💯

PR #310, opened 2022-01-06
Test: fix and add ECS compatibility specs

kares:
somehow specs for `REDISMONLOG` were missing in ECS mode.

also some of the specs adjusted here were run in default mode which on LS 8.x is ECS on by default 
-> yet the specs assumed legacy behavior 

:green_circle: CI


kaisecheng (approved):
LGTM ⭐

PR #311, opened 2022-05-16
Fix: parsing x-edge-location in CLOUDFRONT_ACCESS_LOG

kares:
NOTE: only fixed in ECS mode, legacy mode continues to use `WORD`.

Using `WORD` seem to have worked historically but most edge-locations these days won't match!

AWS CloudFront CDN Edge Locations have a fixed format: https://www.feitsui.com/en/article/3
Thus introduced the `CLOUDFRONT_EDGE_LOCATION` pattern, it's still technically a bug fix.


kaisecheng (approved):
LGTM

Issue #312, opened 2022-06-16
Typo in logstash-patterns-core/patterns/ecs-v1/firewalls


ThomSwiss:
**Logstash information**:

Please include the following information:

1. Logstash version 8.2.2
2. Logstash installation source: official repo of DEB packages

**Description of the problem including expected versus actual behavior**:
- Logstash pipeline crashes
- In the filter logstash-patterns-core/patterns/ecs-v1/firewalls 
 (on my installation located here: ./usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-patterns-core-4.3.3/patterns/ecs-v1/firewalls)

on line 63, CISCOFW302013_302014_302015_302016 you have a typo. Please correct 
**[source][user][name?]**
to 
**[source][user][name]**
When I change this on my installation, it solves the problem. 

**Steps to reproduce**:
Just use the filter with firewall logs

**Provide logs (if relevant)**:
[...]

PR #313, opened 2022-06-17
fix CISCOFW302013_302014_302015_302016 grok pattern

leandrojmp:
This pattern has a wrong reference to the field source.user.name

It is:
```
(?:\(%{DATA:[source][user][name?]}\))
```
And it should be changed to:
```
(?:\(%{DATA:[source][user][name]}\))
```

fix: #312


jsvd:
@leandrojmp thanks for the contribution, I modified the pattern's test to ensure it has a "[source][user][name]", and bumped the patch version to facilitate the release.

jsvd:
Version 4.3.4 has been released with this fix. thanks again!

